PR #1, opened 2025-09-05
Add initial repo template

chtruong814:
Add initial repo template

### DIFF
--- a/.github/workflows/_build_container.yml
+++ b/.github/workflows/_build_container.yml
@@ -72,7 +72,7 @@ jobs:
     with:
       image-name: ${{ inputs.image-name }}
       dockerfile: ${{ inputs.dockerfile }}
-      image-label: nemo-core
+      image-label: ${{ inputs.image-name }}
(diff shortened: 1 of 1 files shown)


.github/workflows/_build_container.yml
```diff
@@ -0,0 +1,83 @@
...
+          echo "$LAST_PRS" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+
+  build:
+    uses: NVIDIA-NeMo/FW-CI-templates/.github/workflows/_build_container.yml@v0.29.0
+    needs: [pre-flight]
+    with:
+      image-name: ${{ inputs.image-name }}
+      dockerfile: ${{ inputs.dockerfile }}
+      image-label: nemo-core
```
thomasdhc: do we want to update this in the cookiecutter?

chtruong814: Yeah let me fix it here and then also update cookiecutter

PR #6, opened 2025-09-16
Add dependencies to pyproject and uv.lock

chtruong814:
Add dependencies to pyproject and uv.lock

* Set uv.lock managed = true to ensure dependency scanning works
* Do not reinstall torch and other dependencies in the container

### DIFF
--- a/docker/Dockerfile.ci
+++ b/docker/Dockerfile.ci
@@ -34,7 +34,6 @@ RUN --mount=type=bind,source=pyproject.toml,target=/workspace/pyproject.toml \
     uv sync --link-mode symlink --locked --all-groups \
         --no-install-package absl-py \
         --no-install-package torch \
-        --no-install-package torchvision \
         --no-install-package triton \
         --no-install-package nvidia-cublas-cu12 \
         --no-install-package nvidia-cuda-cupti-cu12 \


docker/Dockerfile.ci
```diff
@@ -31,5 +31,21 @@ RUN --mount=type=bind,source=pyproject.toml,target=/workspace/pyproject.toml \
 
     # Use the container's torch installation rather than reinstall it
     uv venv --system-site-packages ${UV_PROJECT_ENVIRONMENT}
-    uv sync --link-mode symlink --locked --only-group test --only-group dev
+    uv sync --link-mode symlink --locked --all-groups \
+        --no-install-package absl-py \
+        --no-install-package torch \
+        --no-install-package torchvision \
```
skyw: I'm not familiar with uv. Does this mean install or not install torchvision?
Torchvision is not needed.

chtruong814: It means NOT to install.  I'll remove it anyway because it's not needed here.  I copied these exclusions from sone other projects where we need to prevent uv from overriding the torch dependencies that are already installed.

chtruong814: uv works well when we're installing all the dependencies from scratch. However, in the Pytorch container, we need to tell uv to not install the already installed dependencies we care about

PR #9, opened 2025-09-18
Enable code coverage

chtruong814:
Enable code coverage


skyw (approved):
Approved based on coverage reporting is generated successfully. 

I'm not familiar with workflow syntax, so didn't review the content thoroughly , only skimmed through for obvious.

PR #21, opened 2025-09-25
Add document for soap

skyw:
Plus some small adjustment.

### DIFF
--- a/docs/apidocs/index.md
+++ b/docs/apidocs/index.md
@@ -9,5 +9,5 @@ NeMo Emerging Optimizers API reference provides comprehensive technical document
 utils.md
 orthogonalized-optimizers.md
 soap.md
-scalar_optimizers.md
+scalar-optimizers.md
 ```
--- a/docs/apidocs/index.md
+++ b/docs/apidocs/index.md
@@ -8,4 +8,6 @@ NeMo Emerging Optimizers API reference provides comprehensive technical document
 
 utils.md
 orthogonalized-optimizers.md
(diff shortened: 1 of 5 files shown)


docs/conf.py
```diff
@@ -104,7 +104,7 @@ def linkcode_resolve(domain, info):
         }
     ],
     "switcher": {
-        "json_url": "../versions1.json",
+        "json_url": "versions1.json",
```
aschilling-nv: This needs to remain "../versions1.json." It may throw a minor error during the build, which is fine; the versions1.json file needs to sit outside of the docs folder.

docs/apidocs/index.md
```diff
@@ -8,4 +8,6 @@ NeMo Emerging Optimizers API reference provides comprehensive technical document
 
 utils.md
 orthogonalized-optimizers.md
+soap.md
+scalar_optimizers.md
```
aschilling-nv:
```suggestion
scalar-optimizers.md
```

aschilling-nv: The file name has a "-", which is correct and should match in the TOC.

When possible, always use a "-" instead of a "_" for markdown files.

docs/conf.py
skyw: Rolled back.

docs/apidocs/index.md
skyw: Understood and fixed.. I updated the file name last but forgot to update it in index.md.

PR #30, opened 2025-09-30
save unnecessary matmul

skyw:
Only diagonal of approx_eigenvalue_matrix is needed for _orthogonal_iteration, added logic to skip the last matrix multiply.

Frobenius norm is unitarily invariant, that is for any orthogonal matrix Q and square matrix A ,  $||Q^T A Q||_F = ||A||_F$. So we can calculate norm on kronecker_factor instead of approx_eigenvalue_matrix.


emerging_optimizers/soap/soap_utils.py
```diff
@@ -85,13 +85,9 @@ def get_eigenbasis_eigh(
                 # We use an empty tensor so that the `precondition` function will skip this factor.
                 updated_eigenbasis_list.append(torch.empty(0, device=kronecker_factor.device))
                 continue
-            # Construct approximated eigenvalues using QL^T@L@QL or QR^T@R@QR.
-            # The approximated eigenvalues should be close to diagonal if the eigenbasis is close to the true eigenbasis of the kronecker factor
-            # (i.e. the approximated eigenvectors diagonalize the kronecker factor)
-            approx_eigenvalue_matrix = eigenbasis.T @ kronecker_factor @ eigenbasis
             # Update eigenbasis when necessary. Update is skipped only when adaptive update criteria is met.
             if _adaptive_criteria_met(
```
mkhona-nvidia: `_adaptive_criteria_met` also extracts the diagonal and computes diagonal norm, in addition to matrix frobenius norm. So I think this will need the full approx_eigenvalue_matrix diagonal for diagonal norm and the kronecker factor for frobenius norm, separately.

emerging_optimizers/soap/soap_utils.py
```diff
@@ -316,3 +310,28 @@ def _adaptive_criteria_met(
 
     # check if normalized diagonal component is not smaller than tolerance
     return not utils.eig.adaptive_early_exit_criteria(approx_eigenvalue_matrix, tolerance)
+
+
+def _conjugate(a: torch.Tensor, p: torch.Tensor, diag: bool = False) -> torch.Tensor:
```
mkhona-nvidia, line 292: nit: I am calling all single matrix function inputs `X` throughout the repository

emerging_optimizers/soap/soap_utils.py
skyw: right. Added it back.

emerging_optimizers/soap/soap_utils.py
```diff
@@ -316,3 +310,28 @@ def _adaptive_criteria_met(
...
+    """Calculate similarity transformation
+
+    This function calculates :math:`B = P^T A P`. It assumes P is orthogonal so that :math:`P^{-1} = P^T` and
```
mkhona-nvidia, line 295: Comment has capitalized A and P, while input has lowercare

emerging_optimizers/soap/soap_utils.py
skyw: It was chosen to match math equation. will leave it for the internal functin.

mkhona-nvidia (approved):
New change

PR #32, opened 2025-09-30
Update README.md - Added background and usage example

sbhavani:


README.md
```diff
@@ -6,6 +6,30 @@ Emerging Optimizers is a research project focused on understanding and optimizin
 
 > ⚠️ Note: Emerging-Optimizers is under active development. All APIs are experimental and subject to change. New features, improvements, and documentation updates are released regularly. Your feedback and contributions are welcome, and we encourage you to follow along as new updates roll out.
 
+## Background
+
+### What are Shampoo Optimizers?
+
+Shampoo class optimizers are a family of second-order optimization algorithms that use preconditioned gradient descent to achieve faster convergence compared to traditional first-order methods like Adam or SGD. Unlike these conventional optimizers that treat each parameter independently, Shampoo optimizers leverage the full or block-wise structure of the gradient's second moment statistics through matrix preconditioning.
```
skyw: It is debatable whether they are 2nd order. And not all of optimizers will be supported by emerging optimizers are Shampoo class

README.md
```diff
@@ -6,6 +6,30 @@ Emerging Optimizers is a research project focused on understanding and optimizin
...
+
+**Reference:** [Shampoo: Preconditioned Stochastic Tensor Optimization](https://arxiv.org/abs/1802.09568)
+
+### Why They Matter
+
+Shampoo optimizers have demonstrated significant practical impact in large-scale language model training. Most notably, they were used to train the **Kimi K2 model** ([arXiv:2507.20534](https://arxiv.org/abs/2507.20534)), showcasing their effectiveness at scale. These optimizers can:
```
skyw: Kimi used Muon. And it is also debatable whether it is shampoo.

README.md
```diff
@@ -6,6 +6,30 @@ Emerging Optimizers is a research project focused on understanding and optimizin
...
+
+- Achieve faster convergence, reducing the number of training steps required
+- Improve final model quality through better conditioning of the optimization landscape
+- Enable more efficient hyperparameter tuning due to reduced sensitivity to learning rates
+
+### Optimizers Included
+
+This project focuses on the following Shampoo class optimizers:
```
skyw: More will come.

README.md
```diff
@@ -22,6 +46,55 @@ cd Emerging-Optimizers
...
+        opt.zero_grad()
+    loss = model(batch)
+    loss.backward()
+    for opt in [optimizer_muon, optimizer_adam]:
+        opt.step()
+```
+
+### Integration with Megatron Core
+
+For large-scale training with Megatron Core, see the [integration example](https://github.com/NVIDIA/Megatron-LM/pull/1813) that demonstrates usage with Dense and MoE models.
```
skyw: This can be added when it is integrated into megatron.

README.md
```diff
@@ -22,6 +46,55 @@ cd Emerging-Optimizers
...
 ```
 
-## User guide
+## Usage
+
+### Muon Optimizer
+
+Muon (MomentUm Orthogonalized by Newton-schulz) uses orthogonalization for 2D parameters:
+
+```python
```
skyw: Add this to https://docs.nvidia.com/nemo/emerging-optimizers/latest/. or simply point to a test which show the same usage.

README.md
mkhona-nvidia: Matrix-based preconditioning optimizers might be a better word. Or "non-diagonal" preconditioning

README.md
sbhavani: I like the idea of pointing to an existing test or example, I'll take a look

README.md
sbhavani: I'd still like to keep some pointer here to the PRs we're making in Megatron since it shows the integration. Could I mention that integration is in progress and that this is a PR?

README.md
skyw: We probably should just change to emerging optimizers to not exclude any future ones.

sbhavani: Updated to mention emerging optimizers

README.md
```diff
@@ -2,16 +2,40 @@
...
+
+### Why They Matter
+
+Emerging optimizers have demonstrated significant practical impact in large-scale language model training. Most notably, **Muon was used to train the Kimi K2 model** ([arXiv:2507.20534](https://arxiv.org/abs/2507.20534)), showcasing the effectiveness of these novel approaches at scale. These optimizers can:
+
+- Achieve faster convergence, reducing the number of training steps required
+- Improve final model quality through better conditioning of the optimization landscape
+- Enable more efficient hyperparameter tuning due to reduced sensitivity to learning rates
+
+### Optimizers Included
```
skyw: We don't have original shampoo, also there are a lot more scalar optimizers we also support.
I'd suggest remove this section and let @mkhona-nvidia write it.

PR #39, opened 2025-10-03
Improve tests for scalar optimizers

mkhona-nvidia:
Added flags for scalar optimizer test and improved device handling

### DIFF
--- a/tests/test_scalar_optimizers.py
+++ b/tests/test_scalar_optimizers.py
@@ -44,9 +44,9 @@ class ScalarOptimizerTest(parameterized.TestCase):
         self.device = FLAGS.device
 
     def test_calculate_adam_update_simple(self) -> None:
-        exp_avg_initial = torch.tensor([[1.0]])
-        exp_avg_sq_initial = torch.tensor([[2.0]])
-        grad = torch.tensor([[0.5]])
+        exp_avg_initial = torch.tensor([[1.0]], device=self.device)
+        exp_avg_sq_initial = torch.tensor([[2.0]], device=self.device)
+        grad = torch.tensor([[0.5]], device=self.device)
 
         # Move tensors to the test device
         exp_avg_initial = exp_avg_initial.to(self.device)
@@ -87,7 +87,7 @@ class ScalarOptimizerTest(parameterized.TestCase):
         )
 
         # Manually set Adam's internal state to match conditions *before* the current update
-        adam_optimizer.state[param]["step"] = torch.tensor(float(step - 1))
+        adam_optimizer.state[param]["step"] = torch.tensor(float(step - 1), device=self.device)
         adam_optimizer.state[param]["exp_avg"] = exp_avg_initial.clone()
         adam_optimizer.state[param]["exp_avg_sq"] = exp_avg_sq_initial.clone()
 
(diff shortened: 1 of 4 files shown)


emerging_optimizers/scalar_optimizers/laprop.py
```diff
@@ -76,7 +76,7 @@ def calculate_laprop_update(
 
     # construct the denominator of the inner ADAM optimizer
     second_moment = exp_avg_sq / bias_correction2
-    second_moment = second_moment.sqrt() + eps
+    second_moment.sqrt_().add_(eps)
```
skyw: Don't over use inplace, it doesn't run faster (if not slower) usually, and reduce readability. 
Or it complains device mismatch without the inplace?

tests/test_scalar_optimizers.py
```diff
@@ -23,22 +24,54 @@
 )
 
 
+# Define command line flags
+flags.DEFINE_string("device", "cpu", "Device to run tests on: 'cpu', 'cuda', or 'auto'")
```
skyw: "auto" is over engineered. We'll always control device, or use default. No reason for auto, remove.

tests/test_scalar_optimizers.py
```diff
@@ -23,22 +24,54 @@
...
+flags.DEFINE_integer("seed", 42, "Random seed for reproducible tests")
+flags.DEFINE_boolean("skip_gpu_tests", False, "Skip GPU tests even if CUDA is available")
+
+FLAGS = flags.FLAGS
+
+
 # Base class for tests requiring seeding for determinism
 class BaseTestCase(parameterized.TestCase):
```
skyw: Same as many other places, remove this base class and inherience.

tests/test_scalar_optimizers.py
```diff
@@ -23,22 +24,54 @@
...
+        torch.manual_seed(FLAGS.seed)
         # Set seed for CUDA if available
         if torch.cuda.is_available():
-            torch.cuda.manual_seed_all(42)
+            torch.cuda.manual_seed_all(FLAGS.seed)
+
+        # Set up device based on flags
+        self.device = self._get_test_device()
+
+    def _get_test_device(self):
```
skyw: same as auto, over engineered and hard to read, remove.

tests/test_scalar_optimizers.py
```diff
@@ -23,22 +24,54 @@
...
+        elif FLAGS.device == "cuda":
+            if not torch.cuda.is_available():
+                self.skipTest("CUDA not available")
+            if FLAGS.skip_gpu_tests:
+                self.skipTest("GPU tests skipped by flag")
+            return "cuda"
+        else:
+            return "cpu"
+
+    def _move_to_device(self, *tensors):
```
skyw: count readability, remove.

tests/test_scalar_optimizers.py
```diff
@@ -23,22 +24,54 @@
...
+        """Helper method to move tensors to the test device."""
+        return tuple(tensor.to(self.device) for tensor in tensors)
 
 
 class ScalarOptimizerTest(BaseTestCase):
     def test_calculate_adam_update_simple(self) -> None:
         exp_avg_initial = torch.tensor([[1.0]])
```
skyw: use `torch.tensor([[1.0]], device=FLAGS.devie)`

tests/test_scalar_optimizers.py
```diff
@@ -59,7 +92,7 @@ def test_calculate_adam_update_simple(self) -> None:
             eps=eps,
         )
 
-        initial_param_val_tensor = torch.tensor([[10.0]])
+        initial_param_val_tensor = torch.tensor([[10.0]]).to(self.device)
```
skyw, line 76: same as above, set device when creating tensor to save a memcpy.

tests/test_scalar_optimizers.py
```diff
@@ -23,22 +24,54 @@
...
+        self.device = self._get_test_device()
+
+    def _get_test_device(self):
+        """Get the device to use for testing based on flags."""
+        if FLAGS.device == "auto":
+            return "cuda" if torch.cuda.is_available() and not FLAGS.skip_gpu_tests else "cpu"
+        elif FLAGS.device == "cuda":
+            if not torch.cuda.is_available():
+                self.skipTest("CUDA not available")
+            if FLAGS.skip_gpu_tests:
```
skyw: If all tests in a test case is suppose to run with CUDA, don't skip if no GPU is found, let it fail. Because it means it launched to the wrong machine or machine has issue.
Silent skipping and passing test is very bad behavior.

skyw (changes requested):

emerging_optimizers/scalar_optimizers/laprop.py
mkhona-nvidia: neither, changed back to non-inplace code

skyw (approved):

PR #40, opened 2025-10-03
Fix version issues for MCore container

BoxiangW:
Fix version issues for MCore container


skyw:
Also update readme plz.

PR #42, opened 2025-10-06
use upstream container for ci

skyw:


skyw:
Failed attempt to use upstream container. More setting changes are necessary to make it work.

PR #46, opened 2025-10-08
Feature: skip CI when docs-only label added

pablo-garay:


.github/workflows/cicd-main.yml
```diff
@@ -222,3 +252,13 @@ jobs:
...
+
+  codecov-placeholder:
+    name: codecov/patch
+    needs: [pre-flight]
+    if: needs.pre-flight.outputs.docs_only == 'true'
+    runs-on: ubuntu-latest
+    steps:
+      - name: codecov_placeholder
+        run: |
+          echo "This is a placeholder status check for when no tests are ran but the codecov status is expected"
```
ko3n1g: I would recommend using MBridge's approach instead: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/main/.github/workflows/cicd-main.yml#L181

For GitHub, a status check passes under two conditions: When it passes (trivial) and when it's being skipped. So in the scenarios where we skip the codecov-placeholder, GitHub will accept `codecov/patch` as a valid status check. If the actual job in line 206 failed, the skipped one will take precedence and GitHub will allow merging

pablo-garay: made changes, PTAL

PR #49, opened 2025-10-09
Add option to use triton tsyrk_ex kernels in muon

skyw:


tests/ci/L1_Tests_GPU.sh
```diff
@@ -20,6 +21,6 @@ python tests/test_soap_utils.py
 python tests/soap_smoke_test.py
 python tests/test_scalar_optimizers.py --device=cuda
 python tests/test_spectral_clipping_utils.py
-python tests/test_triton_kernels.py TritonKernelsIntegerInputTest
+python tests/test_triton_kernels.py
```
gdengk: Tests of float dtype will fail. So such failure are allowed in L1 Tests?

emerging_optimizers/triton_kernels/syrk.py
```diff
@@ -316,7 +315,7 @@ def tsyrk_ex(
     Returns:
         Output tensor of shape (N, N)
     """
-
+    assert a.dtype == torch.bfloat16, "Input tensor must be bfloat16"
```
gdengk, line 318: I think this assertion is still needed.

emerging_optimizers/orthogonalized_optimizers/muon_utils.py
```diff
@@ -140,10 +145,17 @@ def newton_schulz(
         # is always in FP32.
         X = X.to(torch.bfloat16)
         logging.log_first_n(logging.INFO, "Using BF16 I/O kernels for Newton-Schulz iteration.", 1)
+        if use_syrk:
+            sm_version = torch.cuda.get_device_capability()
+            if sm_version in ((8, 0), (9, 0), (10, 0), (11, 0)):
```
gdengk, line 150: Feel it's better to be placed in the init function otherwise this is getting called for every param.

PR #57, opened 2025-10-13
Generalize fused weight split

skyw:
Decided to remove split fused parameters logic altogether, because how parameters are fused/stacked together is implementation dependent, it is hard to generalize for everything. 
Instead, now it provides interface to plugin more sophisticated orthogonalize function and let user control how to split.

### DIFF
--- a/emerging_optimizers/psgd/procrustes_step.py
+++ b/emerging_optimizers/psgd/procrustes_step.py
@@ -12,6 +12,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from typing import Literal
+
 import torch
 
 import emerging_optimizers.utils as utils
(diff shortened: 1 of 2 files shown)


emerging_optimizers/orthogonalized_optimizers/orthogonalized_optimizer.py
```diff
@@ -201,15 +195,17 @@ def orthogonalize(self, p: torch.Tensor, grad: torch.Tensor) -> torch.Tensor:
...
+        if self.split_fused and self.is_fused_fn(p):  # type: ignore[misc]
+            logging.log_first_n(logging.INFO, f"split fused parameters with {p.shape} by {self.split_fn}", 1)
+            split_grads = self.split_fn(grad)  # type: ignore[misc]
+
+            assert sum([g.numel() for g in split_grads]) == grad.numel(), "Split grads do not sum to the original grad"
+
+            split_grads_whitened = [self.orthogonalize_fn(g) for g in split_grads]
+            split_grad_scales = [self.scale_factor_fn(g.size(0), g.size(1)) for g in split_grads]
+
+            # TODO(skyw): Revisit whether there are cases that concatenating is not done along dim=0.
```
mkhona-nvidia: `nn.conv1d `(https://docs.pytorch.org/docs/stable/generated/torch.nn.Conv1d.html) has 3d filter and so the output has to be reshaped to 3d

skyw: Valid point, that's also one more reason to let user supply orthogonalize function altogether instead of trying to generalize for everything.

Although conv specifically is a completely different case, all rest code assumes 2d, the scale function for example.

PR #58, opened 2025-10-14
improved test for procrustes step and added third order expansion

mkhona-nvidia:
Added the third order expansion and step size with line search from XL Li.

Improved test for `procrustes_step` after discussion with XL Li

### DIFF
--- a/tests/test_procrustes_step.py
+++ b/tests/test_procrustes_step.py
@@ -163,35 +163,43 @@ class ProcrustesStepTest(parameterized.TestCase):
         # Use amplitude recovery setup to compare convergence speed
         n = 10
         Q_init = torch.randn(n, n, device=self.device, dtype=torch.float32)
-        U, S, Vh = torch.linalg.svd(Q_init)
-        Amplitude = Vh.mH @ torch.diag(S) @ Vh
+        u, s, vh = torch.linalg.svd(Q_init)
+        amplitude = vh.mH @ torch.diag(s) @ vh
 
         # Start from the same initial point for both orders
         Q_order2 = torch.clone(Q_init)
         Q_order3 = torch.clone(Q_init)
 
         max_steps = 200
+        tolerance = 0.01
         err_order2_list = []
         err_order3_list = []
 
-        # Run procrustes steps and track error
-        for _ in range(max_steps):
+        # Track convergence steps directly
+        steps_to_converge_order2 = max_steps  # Default to max_steps if doesn't converge
+        steps_to_converge_order3 = max_steps
+        step_count = 0
+
+        while step_count < max_steps:
             Q_order2 = procrustes_step(Q_order2, order=2, max_step_size=max_step_size)
             Q_order3 = procrustes_step(Q_order3, order=3, max_step_size=max_step_size)
 
-            err_order2 = torch.max(torch.abs(Q_order2 - Amplitude)) / torch.max(torch.abs(Amplitude))
-            err_order3 = torch.max(torch.abs(Q_order3 - Amplitude)) / torch.max(torch.abs(Amplitude))
+            err_order2 = torch.max(torch.abs(Q_order2 - amplitude)) / torch.max(torch.abs(amplitude))
+            err_order3 = torch.max(torch.abs(Q_order3 - amplitude)) / torch.max(torch.abs(amplitude))
 
             err_order2_list.append(err_order2.item())
             err_order3_list.append(err_order3.item())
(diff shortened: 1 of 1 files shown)


skyw:
Could you break it to two PRs, one for improving test, one for new functionality?
Review of test only change can be much lighter and faster.

emerging_optimizers/psgd/procrustes_step.py
```diff
@@ -50,13 +57,23 @@ def procrustes_step(Q: torch.Tensor, max_step_size: float = 0.125, eps: float =
...
-        # clip step size to max_step_size, based on a 2nd order expansion.
-        _step_size = torch.clamp(-tr_RQ / tr_RRQ, min=0, max=max_step_size)
-        # If tr_RRQ >= 0, the quadratic approximation is not concave, we fallback to max_step_size.
-        step_size = torch.where(tr_RRQ < 0, _step_size, max_step_size)
-        # rotate Q as exp(a R) Q ~ (I + a R + a^2 R^2/2) Q with an optimal step size by line search
-        # for 2nd order expansion, only expand exp(a R) to its 2nd term.
-        # Q += step_size * (RQ + 0.5 * step_size * RRQ)
-        Q = torch.add(Q, torch.add(RQ, RRQ, alpha=0.5 * step_size), alpha=step_size)
-
+        if order == 2:
```
skyw, line 60: FIx: The code is not readable with the order argument. It should be possible to make it a loop depend on order.

tests/test_procrustes_step.py
```diff
@@ -155,6 +152,86 @@ def test_preserves_determinant_sign_for_real_matrices(self) -> None:
...
+        (0.0625,),
+        (0.125,),
+    )
+    def test_order3_converges_faster_amplitude_recovery(self, max_step_size: float = 0.0625) -> None:
+        """Test that order 3 converges faster than order 2 in amplitude recovery setting."""
+        # Use amplitude recovery setup to compare convergence speed
+        n = 10
+        Q_init = torch.randn(n, n, device=self.device, dtype=torch.float32)
+        U, S, Vh = torch.linalg.svd(Q_init)
+        Amplitude = Vh.mH @ torch.diag(S) @ Vh
```
skyw: No capital letter in variable names.
Expception can be made to those used in math equations, but not else.

tests/test_procrustes_step.py
```diff
@@ -155,6 +152,86 @@ def test_preserves_determinant_sign_for_real_matrices(self) -> None:
...
+            err_order3 = torch.max(torch.abs(Q_order3 - Amplitude)) / torch.max(torch.abs(Amplitude))
+
+            err_order2_list.append(err_order2.item())
+            err_order3_list.append(err_order3.item())
+
+            # Stop if both have converged
+            if err_order2 < 0.01 and err_order3 < 0.01:
+                break
+
+        # Count steps to convergence for each order
```
skyw: and reason to not record this infor when break out the main loop?
Or use while loop and count steps?

tests/test_procrustes_step.py
```diff
@@ -155,6 +152,86 @@ def test_preserves_determinant_sign_for_real_matrices(self) -> None:
...
+        """Test procrustes_step recovers amplitude of real matrix up to sign ambiguity.
+
+        This is the main functional test for procrustes_step. It must recover the amplitude
+        of a real matrix up to a sign ambiguity with probability 1.
+        """
+        for trial in range(10):
+            n = 10
+            Q = torch.randn(n, n, device=self.device, dtype=torch.float32)
+            U, S, Vh = torch.linalg.svd(Q)
+            Amplitude = Vh.mH @ torch.diag(S) @ Vh
```
skyw: No capital letter in variable name.

skyw (changes requested):
let me take a stab on the main order 2 and 3 function. If nothing better comes up, we can merge as it is.

A few small things need to be fixed also.

emerging_optimizers/psgd/procrustes_step.py
skyw: Tried a few alternatives, some with AI. None of which looked a lot better. let's just keep as it is.

Can change 2nd if to elif also add some empty line to make it easier to read.

tests/test_procrustes_step.py
mkhona-nvidia: just improved test to count steps with while loops

skyw (approved):

PR #60, opened 2025-10-15
PSGD-Kron-Pro(crustes) optimizer implementation 

mkhona-nvidia:
This builds on previous PRs for PSGD's helper functions to make the PSGD-Kron-Pro optimizer

### DIFF
--- a/emerging_optimizers/psgd/psgd.py
+++ b/emerging_optimizers/psgd/psgd.py
@@ -200,9 +200,7 @@ def _update_precond_procrustes(
         q_list: List of Kronecker factors.
         lip_const_list: List of Lipschitz constants for the Kronecker factors.
     """
-    dampened_momentum = exp_avg + (
-        damping_noise_scale + torch.finfo(exp_avg.dtype).eps * exp_avg.abs()
-    ) * torch.randn_like(exp_avg)
+    dampened_momentum = exp_avg + (damping_noise_scale + 1e-7 * exp_avg.abs()) * torch.randn_like(exp_avg)
     pg = apply_preconditioner(q_list, dampened_momentum)
     total_numel = pg.numel()
     updated_q_list: List[torch.Tensor] = []


emerging_optimizers/psgd/psgd_kron_contractions.py
```diff
@@ -24,7 +24,6 @@
 ]
 
 
-@torch.compile  # type: ignore[misc]
```
mkhona-nvidia, line 27: compile was giving graph breaks here, I tried my best to fix it, but gave up. We will have to precompile each param's `partial_contraction`  to use it here

skyw: I don't think we need compile here, it solely depends on tensordot internal.

emerging_optimizers/psgd/psgd.py
```diff
@@ -0,0 +1,318 @@
...
+        precond_init_scale: scale of initial preconditioner values.
+        min_precond_lr: Minimum learning rate for preconditioner learning rate schedule.
+        warmup_steps: Warmup steps for preconditioner learning rate schedule.
+        damping_noise_scale: scale of dampening noise added to gradients.
+        max_update_rms: Clip the update RMS to this value (0 means no clipping).
+    """
+
+    def __init__(
+        self,
+        params: Iterable[torch.nn.parameter.Parameter],
```
skyw: PyTorch optimizer supports more than this. Can use `from torch.optim.optimizer import ParamsT` as type hint.

emerging_optimizers/psgd/psgd.py
```diff
@@ -0,0 +1,318 @@
...
+        lr: float = 3e-3,
+        weight_decay: float = 0.01,
+        use_decoupled_weight_decay: bool = True,
+        momentum: float = 0.9,
+        betaL: float = 0.9,
```
skyw: use beta_lr if it is a learning rate, which is more readable and not violate style.

emerging_optimizers/psgd/psgd.py
```diff
@@ -0,0 +1,318 @@
...
+        precond_lr: float = 0.1,
+        precond_init_scale: float = 1.0,
+        damping_noise_scale: float = 0.1,
+        min_precond_lr: float = 0.3,
+        warmup_steps: int = 10000,
+        max_update_rms: float = 0.0,
+    ) -> None:
+        if betaL is None:
```
skyw: Typehint shows betaL is not Optional, this and the next precond_lr check is contradicting type hint and unnecessary.

emerging_optimizers/psgd/psgd.py
```diff
@@ -0,0 +1,318 @@
...
+        Q: List of Kronecker factors.
+        L: List of Lipschitz constants for the Kronecker factors.
+    """
+    Q: List[torch.Tensor] = []
+    L: List[torch.Tensor] = []
+
+    # Create identity matrices scaled by precond_init_scale for each dimension
+    for size in grad.shape:
+        Q.append(torch.eye(size, device=grad.device) * precond_init_scale)
+        L.append(torch.tensor(1.0, device=grad.device))
```
skyw: Nit: favor torch.ones()

emerging_optimizers/psgd/psgd.py
```diff
@@ -0,0 +1,318 @@
...
+        exp_avg: Exponential moving average of gradient.
+        damping_noise_scale: Scale of noise added to gradient.
+        precond_lr: Learning rate.
+        betaL: Beta for the lower bound.
+
+    Returns:
+        Q: List of Kronecker factors.
+        L: List of Lipschitz constants for the Kronecker factors.
+    """
+    Pg = apply_preconditioner(Q, _dampen_tensor(exp_avg, damping_noise_scale))
```
skyw: Avoid capital letter.

emerging_optimizers/psgd/psgd.py
```diff
@@ -0,0 +1,318 @@
...
+    total_numel = Pg.numel()
+    updated_Q: List[torch.Tensor] = []
+    updated_L: List[torch.Tensor] = []
+    for dim, q in enumerate(Q):
+        # compute gradient covariance
+        precond_grad_cov = partial_contraction(Pg, Pg, dim)
+        if q.dim() < 2:
+            # diagonal or scalar-structured preconditioner
+            q, l_updated = _update_1d_preconditioner(q, L[dim], precond_grad_cov, total_numel, precond_lr, betaL)
```
skyw: Make "update" consistently prefix or suffix, don't use both.

emerging_optimizers/psgd/psgd.py
```diff
@@ -0,0 +1,318 @@
...
+            # matrix-structured preconditioner
+            q, l_updated = _update_matrix_preconditioner(q, L[dim], precond_grad_cov, total_numel, precond_lr, betaL)
+        updated_Q.append(q)
+        updated_L.append(l_updated)
+
+    return updated_Q, updated_L
+
+
+def _update_matrix_preconditioner(
+    q: torch.Tensor,
```
skyw, line 228: Fix: Don't use capital letters for list and lower case for single tensor, Q vs q for example. This is a strong style violation.

it is `kronecker_factor_list` in soap, but here "  Q: List of Kronecker factors.". should use consistent names and avoid capital letter, exception should be rare.
Similarly, lip_const for single tensor, and lip_const_list for list of Lipschitz constants

emerging_optimizers/psgd/psgd.py
```diff
@@ -0,0 +1,318 @@
...
+        precond_grad_cov: Gradient covariance.
+        total_numel: Total number of elements in the gradient.
+        precond_lr: Learning rate.
+        betaL: Beta for the Lipschitz constant exponential moving average.
+
+    Returns:
+        q: Updated Kronecker factor 1D tensor for this dimension.
+        L: Updated Lipschitz constant for this dimension.
+    """
+    normalization = total_numel / q.numel()
```
skyw, line 279: DRY. combine this with _update_matrix_preconditioner.
For example, q.shape[0] == q.numel() if q is 1d. The last line seems have subtle difference, but even a if condition for the last line is better than split the two function with a lot of copied code and comment.

emerging_optimizers/psgd/psgd.py
```diff
@@ -0,0 +1,318 @@
...
+
+    Returns:
+        The preconditioner learning rate.
+    """
+
+    scheduled_lr = precond_lr / math.sqrt(1.0 + step / warmup_steps)
+    return max(scheduled_lr, min_precond_lr)
+
+
+def _dampen_tensor(x: torch.Tensor, scale: float) -> torch.Tensor:
```
skyw: Suggestion: if this function is only called at one place, it is simple enough to be inline than a function.

tests/test_psgd_convergence.py
```diff
@@ -0,0 +1,207 @@
...
+
+            optimizer.step()
+            final_loss = loss.item()
+
+        # Check convergence
+        self.assertLess(final_loss, initial_loss, "Loss should decrease during optimization")
+        self.assertLess(final_loss, 0.01, "Should converge reasonably close to minimum")
+
+        # Check that x is close to target
+        torch.testing.assert_close(x, target, atol=1e-1, rtol=1e-1)
```
skyw, line 101: This is not very close. Consider adding some more comment.

emerging_optimizers/psgd/psgd.py
mkhona-nvidia: it is the ema coefficient for the lipshitz constant, renamed it to beta_lip

emerging_optimizers/psgd/psgd.py
mkhona-nvidia: the `ell` computation is also different, for matrix it uses the lower bound function

lixilinx:
Thanks, @mkhona-nvidia for the PSGD code. It looks good and well organized to me!

I once verified the correctness of psgd_kron_contractions by comparison with einsum. 

In the norm_lower_bound_spd, we will set the default subspace dim to 32 for float32 (based on my test).

emerging_optimizers/psgd/psgd.py
skyw: I think they can be unified. Let's leave it to future improvement.

skyw (dismissed):

mkhona-nvidia:
> Thanks, @mkhona-nvidia for the PSGD code. It looks good and well organized to me!
> 
> I once verified the correctness of psgd_kron_contractions by comparison with einsum.
> 
> In the norm_lower_bound_spd, we will set the default subspace dim to 32 for float32 (based on my test).

The momentum dampening has also been changed to:

### Dampened momentum calculation:
```python
dampened_momentum = exp_avg + (
    damping_noise_scale + torch.finfo(exp_avg.dtype).eps * exp_avg.abs()
) * torch.randn_like(exp_avg)
```

with a `damping_noise_scale` default of 1e-9

emerging_optimizers/psgd/psgd.py
```diff
@@ -0,0 +1,304 @@
...
+        damping_noise_scale: Scale of noise added to gradient.
+        precond_lr: Learning rate.
+        beta_lip: EMA beta for the Lipschitz constant.
+
+    Returns:
+        q_list: List of Kronecker factors.
+        lip_const_list: List of Lipschitz constants for the Kronecker factors.
+    """
+    dampened_momentum = exp_avg + (
+        damping_noise_scale + torch.finfo(exp_avg.dtype).eps * exp_avg.abs()
```
skyw: Fix: torch.finfo(exp_avg.dtype).eps shouldn't be used. Use explicit value. torch.finfo(bfloat16).eps is very large. If the function is never supposed to be used for bfloat16 but only fp32, then even less reason to use eps, which I believe is ULP of 1 in torch.

tests/test_psgd_convergence.py
mkhona-nvidia: change to per element (RMS error) since this is basically matrix fitting with MSE loss

emerging_optimizers/psgd/psgd.py
mkhona-nvidia: changed to 1e-7.

(@lixilinx for vis)

PR #61, opened 2025-10-15
Tensor parallel muon

nagarajankarthik:
Add tensor parallel support for Muon optimizer.


skyw:
Hi, @nagarajankarthik, thanks for the contribution.

We've decided this repo depends on PyTorch only. Training framework (megatron for example) specific implementation should go to that framework.

TP muon is available in megatron, https://github.com/NVIDIA/Megatron-LM/blob/dev/megatron/core/optimizer/muon.py#L43

PR #68, opened 2025-10-31
Update algorithmic choices for soap and clean up code

skyw:
Couple of changes (subject to change further before out of Draft):
* Update is not skipped the first step out of adam warmup
* Current gradient is used in preconditioning
* First Kronecker factor is updated with Shamoo beta
* Explicitly differentiate 0-based and 1-based step count for different purposes.

Also pulled in `https://github.com/nikhilvyas/SOAP` as testing reference. Code is shuffled to match updated algorithmic choices. A version matching unmodified (almost) reference is archived in branch https://github.com/NVIDIA-NeMo/Emerging-Optimizers/tree/skyw/soap_vs_reference_backup_donot_merge

### DIFF
--- a/emerging_optimizers/soap/soap.py
+++ b/emerging_optimizers/soap/soap.py
@@ -273,17 +273,17 @@ class SOAP(optim.Optimizer):
                 torch.cuda.nvtx.range_push("precondition")
                 if state["step"] >= self.adam_warmup_steps:
                     with utils.fp32_matmul_precision(self.fp32_matmul_prec):
-                        norm_precond_update = precondition(
+                        precond_update = precondition(
                             grad=adam_update,
                             eigenbasis_list=state.get("Q", None),
                             dims=[[0], [1]],
                         )
                 else:
-                    norm_precond_update = adam_update
+                    precond_update = adam_update
                 torch.cuda.nvtx.range_pop()
 
-                _clip_update_rms_in_place(norm_precond_update, self.max_update_rms)
-                p.add_(norm_precond_update, alpha=-group["lr"])
+                _clip_update_rms_in_place(precond_update, self.max_update_rms)
+                p.add_(precond_update, alpha=-group["lr"])
 
                 state["step"] += 1
 


emerging_optimizers/soap/soap.py
```diff
@@ -160,155 +161,121 @@ def step(self, closure: Callable[[], float] | None = None) -> float | None:
                 if "step" not in state:
                     state["step"] = 0
 
-                # State initialization
-                # (TODO @mkhona): Better way to check state initialization - use state initializer?
```
mkhona-nvidia, line 164: let's add back the TODO for better state initialization. We need to build it for future checkpointing upgrades

emerging_optimizers/soap/soap.py
```diff
@@ -160,155 +161,121 @@ def step(self, closure: Callable[[], float] | None = None) -> float | None:
...
+                # in the math equation that needs 1-based iteration count.
+                curr_iter_1_based = state["step"] + 1
+
+                if state["step"] == 0:
                     # Exponential moving average of gradient values
                     state["exp_avg"] = torch.zeros_like(grad)
                     # Exponential moving average of squared gradient values
                     state["exp_avg_sq"] = torch.zeros_like(grad)
-
-                if "Q" not in state:
```
mkhona-nvidia, line 171: This can be dangerous in fine-tuning cases where we sometimes dump the optimizer moments or some algorithms which dump the optimizer moment. Depending on SW framework, "step" might not be dumped (as we know from TE/Megatron mismatches)

skyw: It's equally dangerous to rely on any of them. the state_dict load/store of this class will have the right behavior. assumption should not be made on other SW depend on it.

emerging_optimizers/soap/soap.py
skyw: It is not immediately obvious a better one existed. Also no dependency should be made to subsequent SW who overwrite the state dict.

emerging_optimizers/soap/soap.py
```diff
@@ -160,155 +161,121 @@ def step(self, closure: Callable[[], float] | None = None) -> float | None:
...
+                torch.cuda.nvtx.range_pop()
 
-                # Apply weight decay
                 if group["weight_decay"] > 0.0:
                     if self.use_decoupled_wd:
-                        # Apply decoupled weight decay
                         p.add_(p, alpha=(-group["lr"] * group["weight_decay"]))
                     else:
-                        # add l2 regularization before preconditioning (i.e. like adding a squared loss term)
                         grad += group["weight_decay"] * p
```
skyw: @mkhona-nvidia should this happen before precondition?

mkhona-nvidia: yes, it should happen before preconditioning

emerging_optimizers/soap/soap.py
skyw: Added some known improvement under todo.

emerging_optimizers/soap/soap.py
skyw: moved to the correct place

emerging_optimizers/soap/soap.py
skyw: Added an assertion. So at least we know some subclass got things wrong.

emerging_optimizers/soap/soap.py
```diff
@@ -160,155 +161,132 @@ def step(self, closure: Callable[[], float] | None = None) -> float | None:
...
-                            use_adaptive_criteria=self.use_adaptive_criteria,
-                            adaptive_update_tolerance=self.adaptive_update_tolerance,
-                            power_iter_steps=self.power_iter_steps,
+                        norm_precond_update = precondition(
+                            grad=adam_update,
+                            eigenbasis_list=state.get("Q", None),
+                            dims=[[0], [1]],
                         )
+                else:
+                    norm_precond_update = adam_update
```
mkhona-nvidia: We can remove the name "norm" and call it `precond_update` now that normalization of preconditioned gradients has been removed

PR #69, opened 2025-11-03
[DRAFT] added support for 3d weight params arising from conv1d

mkhona-nvidia:


emerging_optimizers/orthogonalized_optimizers/muon_utils.py
```diff
@@ -196,15 +196,35 @@ def newton_schulz_tp(
...
         mode: The mode to use for the Newton-Schulz iteration.
     """
+
+    # Placeholder for 3D conv1d case
+    original_3d_shape = None
+
     if partition_dim is None:
         # Fallback path for non TP params.
+        # Handle 3D conv1d case
+        if x.dim() == 3:
```
skyw, line 206: This is not the right place to add this logic.
the function still handles 2d input, reshape logic should be outside of this function. probably should be in the Megatron inherited muon

mkhona-nvidia: The issue is the same as unfusing QKV: in the optimizer state, there param shape will always be 3D. We would need to add a handle in the `OrthogonalizedOptimizer` class

skyw: no, all of our functions handles 2d.
Logic of making a 3d (or nd for that matter) 2d should be out of it.

PR #70, opened 2025-11-04
Scion optimizer 

mkhona-nvidia:
Added Scion.

The main change is the parametrization to allow Franke-Wolfe. We now *DO NOT* use weight decay and instead use step size (i.e. the learning rate) and spectral radius, with the unit_rms choice for width scaling

### DIFF
--- a/emerging_optimizers/orthogonalized_optimizers/scion.py
+++ b/emerging_optimizers/orthogonalized_optimizers/scion.py
@@ -22,13 +22,6 @@ from emerging_optimizers.orthogonalized_optimizers.muon_utils import newton_schu
 from emerging_optimizers.orthogonalized_optimizers.orthogonalized_optimizer import OrthogonalizedOptimizer
 
 
-_args_doc = """params: Iterable of parameters to optimize or dicts defining parameter groups
-        lr: The learning rate used by the internal SGD.
-        momentum_beta: The momentum used by the internal SGD.
-        spectral_radius: The spectral radius to use for the update, we are scaling the LMO by this spectral radius.
-"""
-
-
 class Scion(OrthogonalizedOptimizer):
     """Scion: Stochastic CondItional descent with Operator Norms
 
@@ -50,8 +43,12 @@ class Scion(OrthogonalizedOptimizer):
           parameters; those should all be optimized by the appropriate LMO for that layer. For example,
           for 1d params, it is scaled by the `ell_inf` radius.
 
+
     Args:
-        {_args_doc}
+        params: Iterable of parameters to optimize or dicts defining parameter groups
+        lr: The learning rate used by the internal SGD.
+        momentum_beta: The momentum used by the internal SGD.
+        fp32_matmul_prec: Precision of the matmul operations in optimizer states GEMM operations.
         coefficient_type: The type of coefficient set to use for the Newton-Schulz iteration. Can be one of
             ["simple", "quintic", "polar_express"].
         num_ns_steps: The number of iteration steps to use in the Newton-Schulz iteration.
@@ -81,7 +78,6 @@ class Scion(OrthogonalizedOptimizer):
         logging.info("Scion does not use weight decay. Setting use_independent_wd to False to allow Franke-Wolfe.")
         use_independent_wd = False
 
-        # Scion does not use Nesterov momentum.
         logging.info("Scion does not use Nesterov momentum. Setting use_nesterov to False.")
         use_nesterov = False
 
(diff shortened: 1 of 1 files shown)


emerging_optimizers/orthogonalized_optimizers/scion.py
```diff
@@ -0,0 +1,124 @@
...
+                logging.error("Triton 3.4.0 or higher is required for use_syrk to be True.")
+                use_syrk = False
+            elif sm_version not in ((8, 0), (9, 0), (10, 0), (10, 3)):
+                logging.error(
+                    f"Correctness of Triton kernel on SM {sm_version} cannot be guaranteed. Setting use_syrk to False."
+                )
+                use_syrk = False
+
+        # Add checks for weight decay arguments to enable Franke-Wolfe step.
+        if weight_decay != 1:
```
skyw: They should be removed in argument, keep the interface simpler.

emerging_optimizers/orthogonalized_optimizers/scion.py
```diff
@@ -0,0 +1,115 @@
...
+
+        logging.warning("Scion does not use weight decay. Setting use_independent_wd to False to allow Franke-Wolfe.")
+        use_independent_wd = False
+
+        def scaled_orthogonalize_fn(grad: torch.Tensor) -> torch.Tensor:
+            logging.debug(
+                f"Orthogonalizing grad with {num_ns_steps} steps, {coefficient_type} coefficient, spectral_radius={spectral_radius}"
+            )
+            orth_grad = newton_schulz(grad, steps=num_ns_steps, coefficient_type=coefficient_type, use_syrk=use_syrk)
+            width_factor = (grad.size(-2) / grad.size(-1)) ** 0.5
```
skyw: Also isn't this spectral scaling? If yes, can just use "get_muon_scale_factor"

emerging_optimizers/orthogonalized_optimizers/scion.py
mkhona-nvidia: done, removed from args and added defaults in `init`

emerging_optimizers/orthogonalized_optimizers/scion.py
mkhona-nvidia: We don't want to specify options here, Scion mathematically needs this scaling only and nothing else

skyw: you don't need any option, just call that function with fixed argument

mkhona-nvidia: done

emerging_optimizers/orthogonalized_optimizers/scion.py
```diff
@@ -0,0 +1,119 @@
...
+          [`arXiv:2502.07529 <https://arxiv.org/abs/2502.07529>`_]
+
+    Warning:
+        - This optimizer requires that all parameters passed in are 2D.
+        - It should not be used for the embedding layer, the final fully connected layer, or any 1-D
+          parameters; those should all be optimized by the appropriate LMO for that layer. For example,
+          for 1d params, it is scaled by the `ell_inf` radius.
+
+    Args:
+        {_args_doc}
```
skyw: This includes argument Scion class doesn't have. Doc should match the class.

emerging_optimizers/orthogonalized_optimizers/scion.py
```diff
@@ -0,0 +1,119 @@
...
+                logging.error("Triton 3.4.0 or higher is required for use_syrk to be True.")
+                use_syrk = False
+            elif sm_version not in ((8, 0), (9, 0), (10, 0), (10, 3)):
+                logging.error(
+                    f"Correctness of Triton kernel on SM {sm_version} cannot be guaranteed. Setting use_syrk to False."
+                )
+                use_syrk = False
+
+        # Add checks for weight decay arguments to enable Franke-Wolfe step.
+        logging.warning("Scion does not use weight decay. Setting weight_decay to 1.")
```
skyw: Change it to info level. Users choose to use this class should know the implication, so it is not a warning to them but infor.

emerging_optimizers/orthogonalized_optimizers/scion.py
```diff
@@ -0,0 +1,119 @@
...
+        fp32_matmul_prec: str = "medium",
+        coefficient_type: str = "quintic",
+        num_ns_steps: int = 5,
+        spectral_radius: float = 1.0,
+        use_syrk: bool = False,
+    ) -> None:
+        if num_ns_steps < 1:
+            raise ValueError(f"num_ns_steps must be at least 1, got {num_ns_steps}")
+
+        if use_syrk:
```
skyw: Recommend to leave out syrk option for this demonstration class for now.

skyw (changes requested):
Some minor changes needed. otherwise LGTM

skyw (approved):

PR #71, opened 2025-11-05
Pre first release cleanup

skyw:
* Slightly updated test.
* Force keyword arguments for optimizers that have very long argument list, otherwise error prone
* Abstract weight decay to single mixin class
* Some code style fix/improvement

@mkhona-nvidia @FDecaYed please take a look, there will be another doc only clean up before we making a release, after which things changing things would become maintenance burden.

@mkhona-nvidia coverage of scalar optimizers and soap with adaptative criteria is low, would be good to improve. It can also be done after release, so not mandatory.

### DIFF
--- a/emerging_optimizers/orthogonalized_optimizers/scion.py
+++ b/emerging_optimizers/orthogonalized_optimizers/scion.py
@@ -70,9 +70,11 @@ class Scion(OrthogonalizedOptimizer):
             raise ValueError(f"num_ns_steps must be at least 1, got {num_ns_steps}")
 
         # Add checks for weight decay arguments to enable Franke-Wolfe step.
-        logging.info("Scion does not use weight decay. Setting weight_decay to 1 and weight_decay_method to l2.")
+        logging.info(
+            "Scion does not use weight decay. Setting weight_decay to 1 and weight_decay_method to decoupled."
+        )
         weight_decay = 1
-        weight_decay_method = "l2"
+        weight_decay_method = "decoupled"
 
         logging.info("Scion does not use Nesterov momentum. Setting use_nesterov to False.")
         use_nesterov = False


emerging_optimizers/orthogonalized_optimizers/orthogonalized_optimizer.py
```diff
@@ -26,23 +26,23 @@
...
-            default to be False.
+        use_nesterov: Whether to use Nesterov-style momentum in the internal SGD.
+        weight_decay_method: Method to apply weight decay, see :class:`~emerging_optimizers.mixin.WeightDecayMixin`
+            for more details.
         fp32_matmul_prec: Precision of the matmul operations in optimizer states GEMM operations.
 """
 
 
-class OrthogonalizedOptimizer(optim.Optimizer):
+class OrthogonalizedOptimizer(opt_mixin.WeightDecayMixin, optim.Optimizer):
```
mkhona-nvidia, line 45: Double inheritance makes this complicated

emerging_optimizers/mixin.py
```diff
@@ -0,0 +1,52 @@
...
+
+from typing import Literal
+
+import torch
+
+
+WeightDecayT = Literal["decoupled", "independent", "l2"]
+
+
+class WeightDecayMixin:
```
mkhona-nvidia, line 24: What is the reason for this to be a class rather than a set of functions that are chosen based on arguments?

emerging_optimizers/orthogonalized_optimizers/orthogonalized_optimizer.py
skyw: No, this is mixin. The only type of multi inheritance should be used in Python.

emerging_optimizers/mixin.py
skyw: weight decay is a function highly coupled with optimizer, and shared among a lot of optim subclass.

Thought about a optimizer based class with this function and make everyone inherit from it, but not all of our optimizers use the same base.

emerging_optimizers/orthogonalized_optimizers/scion.py
```diff
@@ -69,14 +70,9 @@ def __init__(
             raise ValueError(f"num_ns_steps must be at least 1, got {num_ns_steps}")
 
         # Add checks for weight decay arguments to enable Franke-Wolfe step.
-        logging.info("Scion does not use weight decay. Setting weight_decay to 1.")
+        logging.info("Scion does not use weight decay. Setting weight_decay to 1 and weight_decay_method to l2.")
```
mkhona-nvidia: it should be `decoupled` for scion

skyw: oops...fixed.

FDecaYed (commented):
LGTM

PR #76, opened 2025-11-13
Adaptive learning rate for Muon: NorMuon and AdaMuon

mkhona-nvidia:
This MR extends Muon to support adaptive learning rates for Muon-like optimizers, namely NorMuon (https://arxiv.org/pdf/2510.05491) and AdaMuon (https://arxiv.org/pdf/2507.11005).

Normuon recently was incorporated into the speedrun record (https://github.com/KellerJordan/modded-nanogpt/blob/master/train_gpt.py#L595)

### DIFF
--- a/tests/test_adaptive_muon.py
+++ b/tests/test_adaptive_muon.py
@@ -97,20 +97,24 @@ class AdaptiveMuonTest(parameterized.TestCase):
     def test_unknown_moment2_method_raise_type_error(self) -> None:
         """Test that AdaptiveMuon raises TypeError for unknown moment2_method."""
         test_param = nn.Parameter(torch.randint(-5, 5, (8, 16), dtype=torch.float32, device=FLAGS.device))
+        test_param.grad = torch.randint_like(test_param, -5, 5)
 
+        adaptive_opt = AdaptiveMuon(
+            [test_param],
+            lr=0.01,
+            momentum_beta=0.9,
+            weight_decay=0.0,
+            use_nesterov=False,
+            moment2_method=None,
+            beta2=0.999,
+            eps=1e-8,
+            weight_decay_method="decoupled",
+            fp32_matmul_prec="highest",
+        )
+
+        # TypeError is raised during step() when initializing moment2_buffer
         with self.assertRaises(TypeError):
-            AdaptiveMuon(
-                [test_param],
-                lr=0.01,
-                momentum_beta=0.9,
-                weight_decay=0.0,
-                use_nesterov=False,
-                moment2_method=None,
-                beta2=0.999,
-                eps=1e-8,
-                weight_decay_method="decoupled",
-                fp32_matmul_prec="highest",
-            )
+            adaptive_opt.step()
 
 
 if __name__ == "__main__":


emerging_optimizers/orthogonalized_optimizers/adaptive_orthogonalized_optimizer.py
```diff
@@ -0,0 +1,248 @@
...
+        correct_bias: Whether to apply bias correction to second moment.
+        second_moment_method: Method to apply second moment, see :class:`~emerging_optimizers.mixin.SecondMomentMixin`
+            for more details. Options: "adamuon" (elementwise like AdamW), "normuon" (row/column-wise).
+        weight_decay_method: Method to apply weight decay, see :class:`~emerging_optimizers.mixin.WeightDecayMixin`
+            for more details.
+        fp32_matmul_prec: Precision of the matmul operations in optimizer states GEMM operations.
+"""
+
+
+class AdaptiveOrthogonalizedOptimizer(
```
mkhona-nvidia: @skyw I've made a separate class, but the `SecondMomentMixin` can be put into `OrthogonalizedOptimizer` and made into no-ops.

emerging_optimizers/mixin.py
```diff
@@ -51,3 +52,101 @@ def _apply_weight_decay_inplace(
...
+                # Determine which dimension to reduce based on parameter shape
+                avg_dim = -1 if grad.shape[-2] >= grad.shape[-1] else -2
+                # Specify the shape with reduced dimension
+                second_moment_shape = list(grad.shape)
+                second_moment_shape[avg_dim] = 1
+                second_moment = torch.zeros(second_moment_shape, dtype=grad.dtype, device=grad.device)
+            else:
+                raise ValueError(f"Invalid second moment method: {second_moment_method}")
+
+            state["second_moment_buffer"] = second_moment
```
skyw: Rule of thumb is mixin shouldn't allocating anything, nor adding attribute to the class (there are exceptions, but would not apply here).

emerging_optimizers/orthogonalized_optimizers/orthogonalized_optimizer.py
```diff
@@ -102,7 +114,10 @@ def __init__(
         weight_decay: float,
         *,
         use_nesterov: bool,
-        weight_decay_method: opt_mixin.WeightDecayT,
+        second_moment_method: opt_mixin.SecondMomentT = None,
```
skyw: This adding too many things to one class we just cleaned up a bit. Need to consider better design.

skyw (changes requested):
How many will need this 2nd momentum other than OrthogonalizedOptimizer? Looks like none (or very few) because Adam based optimizer won't use it. In which case, should inherit from OrthogonalizedOptimizer and override step() function. 

The choices are between copying code and keeping adding more functionalities to one class. In this case, I think copying some code in step() is the best trade off.

One another need to consider is if it turned out to be not very useful, how painful to remove the code, which would also lead to  inheriting and override step function.

emerging_optimizers/mixin.py
mkhona-nvidia: Mixin removed and `OrthogonalizedOptimizer` subclassed

emerging_optimizers/orthogonalized_optimizers/adaptive_orthogonalized_optimizer.py
```diff
@@ -0,0 +1,265 @@
...
+        beta2: The exponential decay rate for second moment (like AdamW's beta2).
+        eps: Small constant for numerical stability in second moment normalization.
+        weight_decay_method: Method to apply weight decay, see :class:`~emerging_optimizers.mixin.WeightDecayMixin`
+            for more details.
+        fp32_matmul_prec: Precision of the matmul operations in optimizer states GEMM operations.
+"""
+
+
+class AdaptiveOrthogonalizedOptimizer(OrthogonalizedOptimizer):
+    """Orthogonalized optimizer with adaptive second moment (AdaMuon/NorMuon variants).
```
skyw: It this meant to be a base class like OrthogonalizedOptimizer or a variant of Muon? Muon specific arguments don't belong to base class. Inherit from Muon if it meant to be a subclass of muon

emerging_optimizers/orthogonalized_optimizers/adaptive_orthogonalized_optimizer.py
```diff
@@ -0,0 +1,265 @@
...
+
+    This class extends OrthogonalizedOptimizer by adding AdamW-style or NorMuon-style second moment
+    accumulation after orthogonalization. The step() method is overridden to include second moment
+    normalization logic.
+
+    Args:
+        {_adaptive_args_doc}
```
skyw: No class reuse this, no need to create another one. Doing this only when you find same strings are copied too many times.

emerging_optimizers/orthogonalized_optimizers/adaptive_orthogonalized_optimizer.py
```diff
@@ -0,0 +1,265 @@
...
+            momentum_beta,
+            use_nesterov=use_nesterov,
+            weight_decay=weight_decay,
+            weight_decay_method=weight_decay_method,
+            fp32_matmul_prec=fp32_matmul_prec,
+            scaled_orthogonalize_fn=scaled_orthogonalize_fn,
+        )
+
+        for group in self.param_groups:
+            group["beta2"] = beta2
```
skyw: Any reason to not put this in default dict?

emerging_optimizers/orthogonalized_optimizers/adaptive_orthogonalized_optimizer.py
```diff
@@ -0,0 +1,265 @@
...
+            group["eps"] = eps
+
+    def _initialize_second_moment(
```
skyw: Suggestion: I prefer moment2 to be consistent with beta2

emerging_optimizers/orthogonalized_optimizers/adaptive_orthogonalized_optimizer.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def __init__(
+        self,
+        params: ParamsT,
+        lr: float,
+        momentum_beta: float,
+        weight_decay: float,
+        *,
+        use_nesterov: bool,
+        second_moment_method: Literal["adamuon", "normuon"],
```
skyw: Append new argument to the right most be default, unless there is a strong reason to add to the middle.

tests/test_orthogonalized_optimizer.py
```diff
@@ -249,5 +252,91 @@ def test_smoke(self, shape) -> None:
...
+            # Full elementwise buffer
+            self.assertEqual(second_moment.shape, test_param.shape)
+        elif second_moment_method == "normuon":
+            # Reduced shape buffer
+            avg_dim = -1 if shape[-2] >= shape[-1] else -2
+            expected_shape = list(shape)
+            expected_shape[avg_dim] = 1
+            self.assertEqual(list(second_moment.shape), expected_shape)
+
+    def test_requires_second_moment_method(self) -> None:
```
skyw: test name can be more specific, like test_unknwon_moment2_method_raise_error

tests/test_orthogonalized_optimizer.py
```diff
@@ -249,5 +252,91 @@ def test_smoke(self, shape) -> None:
...
+            weight_decay_method="decoupled",
+            fp32_matmul_prec="highest",
+        )
+        adaptive_opt.step()
+
+    @parameterized.parameters(
+        {"shape": (8, 16), "second_moment_method": "adamuon"},
+        {"shape": (16, 8), "second_moment_method": "normuon"},
+    )
+    def test_second_moment_initialization(self, shape, second_moment_method) -> None:
```
skyw: It is better to state what behavior is tested in name, like test_xxx_correct_shape.

Test is also a bit weak, is there any code we can use as testing reference to compare value?

emerging_optimizers/orthogonalized_optimizers/adaptive_orthogonalized_optimizer.py
mkhona-nvidia: It's supposed to be inherited from orthogonalizedoptimizer, so at the same level of abstraction as Muon. It reuses muon's newton-schulz so that's why some arguments are shared. it's not a subclass of muon

tests/test_orthogonalized_optimizer.py
mkhona-nvidia: no code exists, except speedrun  (which this is based on)

emerging_optimizers/orthogonalized_optimizers/adaptive_orthogonalized_optimizer.py
skyw: It is less about algorithm or name but code structure. This is obviously more of subclass of muon given amount of copied code from muon's init and the similarity of argument list.
The point of OrthogonalizedOptimizer is to be able to have different orthogonalization function, this one doesn't have that capability. I would just call it AdaptiveMuon, it is still momentum orthogonalized by NS, MONxxx MONSxxx would also work.

mkhona-nvidia: Makes sense, I changed it to inherit from Muon instead of `OrthogonalizedOptimizer`

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def __init__(
+        self,
+        params: ParamsT,
+        lr: float,
+        momentum_beta: float,
+        weight_decay: float,
+        *,
+        use_nesterov: bool,
+        moment2_method: Literal["adamuon", "normuon"],
```
skyw: If possible, it is better to append arguments to base class's list instead of insert in the middle. Appending more argument to base is much easier to read.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -0,0 +1,242 @@
...
+        eps: float = 1e-8,
+        weight_decay_method: opt_mixin.WeightDecayT = "decoupled",
+        fp32_matmul_prec: str,
+        coefficient_type: str = "quintic",
+        num_ns_steps: int = 5,
+        scale_mode: str = "spectral",
+        extra_scale_factor: float = 1.0,
+        use_syrk: bool = False,
+    ):
+        self.moment2_method = moment2_method
```
skyw: Move this after calling super's init.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -0,0 +1,242 @@
...
+            weight_decay_method=weight_decay_method,
+            fp32_matmul_prec=fp32_matmul_prec,
+            coefficient_type=coefficient_type,
+            num_ns_steps=num_ns_steps,
+            scale_mode=scale_mode,
+            extra_scale_factor=extra_scale_factor,
+            use_syrk=use_syrk,
+        )
+
+        # Add beta2 and eps to param_groups defaults
```
skyw: Remove meaningless comment

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -0,0 +1,242 @@
...
+            step_size = moment2.clamp_min(eps).rsqrt_()
+            return orth_grad * step_size
+
+        else:
+            raise TypeError(f"Invalid second moment method: {self.moment2_method}")
+
+    @torch.no_grad()  # type: ignore[misc]
+    @override
+    def step(self, closure: Callable[[], float] | None = None) -> float | None:
+        """Performs a single optimization step with second moment normalization.
```
skyw: 3rd party singular is not necessary.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -0,0 +1,242 @@
...
+            closure: A closure that reevaluates the model and returns the loss.
+        """
+        if closure is None:
+            loss = None
+        else:
+            loss = closure()
+
+        for group in self.param_groups:
+            for p in group["params"]:
+                if p.dim() == 1:
```
skyw: I know we did this in base class, but I think we should check dim == 2. nd is also not supported.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -0,0 +1,242 @@
...
+                # include nesterov momentum
+                if self.use_nesterov:
+                    grad = grad.lerp(exp_avg, group["momentum_beta"])
+                else:
+                    grad = exp_avg
+
+                with utils.fp32_matmul_precision(self.fp32_matmul_prec):
+                    grad = self.scaled_orthogonalize_fn(grad)
+
+                # Apply second moment normalization
```
skyw: Remove meaningless comment. You see the code literarily says the same thing. clean code speaks for itself and doesn't need comment.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -0,0 +1,242 @@
...
+                grad = self._apply_moment2_normalization(
+                    orth_grad=grad,
```
skyw: It is better to change output of line 228 to orth_grad, otherwise it is confusing here.

We've been sloppy in couple of other places. If it is trivial, can update altogether in this PR. or address in following PRs.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -0,0 +1,242 @@
...
+                    moment2=state["moment2_buffer"],
+                    beta2=group["beta2"],
+                    eps=group["eps"],
+                )
+
+                # perform weight update
+                # scale is applied to have update RMS == 1
```
skyw, line 236: Q: is "# scale is applied to have update RMS == 1" still relevant

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -0,0 +1,242 @@
...
+            fp32_matmul_prec=fp32_matmul_prec,
+            coefficient_type=coefficient_type,
+            num_ns_steps=num_ns_steps,
+            scale_mode=scale_mode,
+            extra_scale_factor=extra_scale_factor,
+            use_syrk=use_syrk,
+        )
+
+        # Add beta2 and eps to param_groups defaults
+        for group in self.param_groups:
```
mkhona-nvidia, line 94: I am adding the extra argument using this. Is there a better way @skyw

skyw: This should be fine, can revisit later.

tests/test_adaptive_orthogonalized_optimizer.py
```diff
@@ -0,0 +1,120 @@
...
+import torch.nn as nn
+from absl import flags
+from absl.testing import absltest, parameterized
+
+from emerging_optimizers.orthogonalized_optimizers.adaptive_muon import (
+    AdaptiveMuon,
+)
+
+
+# Define command line flags
```
skyw: This is also meanless

skyw (dismissed):

skyw (approved):

PR #78, opened 2025-11-21
Add utilities for conv1d support

skyw:
conv1d is used in mamba mixer

### DIFF
--- a/emerging_optimizers/utils/modules.py
+++ b/emerging_optimizers/utils/modules.py
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import math
 from typing import Any, Self
 
-import numpy as np
 import torch
 import torch.nn as nn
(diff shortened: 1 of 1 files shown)


mkhona-nvidia (approved):

tests/test_utils_modules.py
```diff
@@ -0,0 +1,96 @@
...
+# limitations under the License.
+
+import torch
+import torch.nn as nn
+from absl.testing import absltest, parameterized
+
+from emerging_optimizers.utils.modules import Conv1dFlatWeights
+
+
+class TestConv1dFlatWeights(parameterized.TestCase):
```
mkhona-nvidia, line 23: One thing that would be good to test is how this checkpoints relative to regular conv1d (is bias and weight fused)

skyw: Yes, that's just pytorch nn.Module and Parameter behavior

PR #82, opened 2025-12-09
Add Mop

skyw:
Add experimental Momentum Orthogonalized by Polar decomposition.

Polar decomposition is calculated by SVD instead of QDWH

### DIFF
--- a/emerging_optimizers/orthogonalized_optimizers/mop.py
+++ b/emerging_optimizers/orthogonalized_optimizers/mop.py
@@ -75,13 +75,24 @@ MOP.__doc__ = MOP.__doc__.format(_args_doc=_args_doc)  # type: ignore[union-attr
 
 
 def polar_via_svd(A: torch.Tensor, return_p: bool = False) -> tuple[torch.Tensor, Optional[torch.Tensor]]:
-    """Compute polar decomposition via SVD"""
+    """Compute polar decomposition via SVD
 
+    Args:
+        A: The input tensor to compute the polar decomposition of.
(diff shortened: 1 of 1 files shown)


emerging_optimizers/orthogonalized_optimizers/mop.py
```diff
@@ -0,0 +1,87 @@
...
+            weight_decay_method=weight_decay_method,
+            fp32_matmul_prec=fp32_matmul_prec,
+            scaled_orthogonalize_fn=scaled_orthogonalize_fn,
+        )
+
+
+MOP.__doc__ = MOP.__doc__.format(_args_doc=_args_doc)  # type: ignore[union-attr]
+
+
+def polar_via_svd(A: torch.Tensor, return_p: bool = False) -> tuple[torch.Tensor, Optional[torch.Tensor]]:
```
mkhona-nvidia, line 77: what is the functionality of `return_p`, right now the `scaled_orthogonalize_fn` does not do anything with it

Suggestion: add a argument to `scaled_orthogonalize_fn` to additionally add a scale factor of `sqrt(trace(H)) `(this is the polargrad optimizer). Also "p" is a confusing name, since what we are returning is "H".

skyw: Forgot to rename H to p. The decision to use p instead of H is because scipy uses it https://docs.scipy.org/doc//scipy-1.16.2/reference/generated/scipy.linalg.polar.html

return_p argument is to save computation of p as we don't need it, while keep polar_via_svd a function actually capable of returning polar decomposition.

PR #87, opened 2025-12-24
Move 2d shape check to orthogonalize()

skyw:
This will allow subclass to override orthogonalize() and support N-D parameters as it see fit.


FDecaYed (approved):
Changes look good to me

PR #94, opened 2026-02-05
Generalize Newton-Schulz coefficient iterator

### LINKED ISSUE #90: Extend Polar Express to allow arbitrary number of steps
**Is your feature request related to a problem? Please describe.**

What is called "Polar Express" in `muon_utils.py` does not exactly match Algorithm 1 in the _Polar Express_ paper ([APMG25](https://arxiv.org/pdf/2505.16932); see screenshot below). In particular, the current NVIDIA implementation requires `steps` to be divisible by 8, and rotates cyclically through 8 hardcoded quintics. In the Polar Express paper, after step 8, all subsequent steps use the same quintic, equalling the 8th one, up to a small rescaling (the "safety factor"). It appears that NVIDIA's implementation only matches the published algorithm in the case `steps==8`.
 
**Describe the solution you'd like**

Follow the published algorithm for arbitrary `steps`.

**Describe alternatives you've considered**

* Emit a warning when `steps != 8` to alert the user that NVIDIA's implementation diverges from the published reference.
* Implement the more general Polar Express framework as detailed in Algorithm 2 of the paper.

**Additional context**

Image

skyw:
Support repeat last and cycle through coefficient set.

Update "polar_express" coefficient set behavior to match the paper, fix #90

### DIFF
--- a/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
+++ b/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
@@ -21,7 +21,7 @@ from absl import logging
 from emerging_optimizers import triton_kernels
 
 
-__all__ = ["newton_schulz", "newton_schulz_tp", "NSCoeffT", "iter_coefficient_set", "CoeffIterMode"]
+__all__ = ["newton_schulz", "newton_schulz_tp", "NSCoeffT", "get_coefficient_iterator", "CoeffIterMode"]
 
 CoeffIterMode = Literal["cycle", "repeat_last"]
 


mkhona-nvidia (approved):

emerging_optimizers/orthogonalized_optimizers/muon_utils.py
```diff
@@ -12,19 +12,23 @@
...
 
-__all__ = ["newton_schulz", "newton_schulz_tp", "NSCoeffT"]
+__all__ = ["newton_schulz", "newton_schulz_tp", "NSCoeffT", "get_coefficient_iterator", "CoeffIterMode"]
+
+CoeffIterMode = Literal["cycle", "repeat_last"]
 
 NSCoeffT = Literal["simple", "quintic", "polar_express", "aol", "custom"]
 
 _COEFFICIENT_SETS = {
+    # Values are rounded to closest representable in single precision.
```
nick-knight, line 31: I would like to challenge this claim, but without providing evidence. Generally, FP32 constants require up to [9 decimal digits](https://en.cppreference.com/w/cpp/types/numeric_limits/max_digits10.html) to survive the FP32 -> decimal literal -> FP32 round-trip. Now, it's entirely possible that all of the literals in this file are quite special, only needing 4 digits after the decimal point, but it seems highly unlikely to me, given the methods used to derive these constants (Remez, etc.). Rather, it seems much more likely to me that someone simply printed them with a `%.4f` format string.

My suggestion is to delete this comment.

I've been meaning to file #96 for a while, but this prompted me to do so.

nick-knight (approved):
LGTM. Addresses my concern in #90.

Minor concern about a comment added in this PR; forked discussion to #96.

emerging_optimizers/orthogonalized_optimizers/muon_utils.py
skyw: " it's entirely possible that all of the literals in this file are quite special"
yes, @mkhona-nvidia went through bit pattern of all the values and kept enough digits to represent the single precision value.
Most of the values in the publications are not representable in single precision.

PR #107, opened 2026-02-20
Norm-Based Adaptive Moment Estimation with Orthogonalized Momentum

mkhona-nvidia:
Build Namo as another method to normalize Muon updates 

From Adam improves Muon (https://arxiv.org/abs/2602.17080)


emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -70,7 +73,7 @@ def __init__(
         scale_mode: muon.MuonScaleT = "spectral",
         extra_scale_factor: float = 1.0,
         use_syrk: bool = False,
-        moment2_method: Literal["adamuon", "normuon"] = "adamuon",
+        moment2_method: Literal["adamuon", "normuon", "namo"] = "adamuon",
```
skyw: Nit: Define a type since now it has three variant. Can also do it when we have more.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -121,6 +125,9 @@ def _initialize_moment2(
                 moment2_shape = list(grad.shape)
                 moment2_shape[avg_dim] = 1
                 moment2 = torch.zeros(moment2_shape, dtype=grad.dtype, device=grad.device)
+            elif self.moment2_method == "namo":
+                # Scalar second moment: EMA of ||G_t||_F^2
+                moment2 = torch.zeros((), dtype=grad.dtype, device=grad.device)
```
skyw: Q: Why empty tensor? seems a meaningless placeholder.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -132,22 +139,37 @@ def _apply_moment2_normalization(
         moment2: torch.Tensor,
         beta2: float,
         eps: float,
+        *,
+        grad_fro_sq: torch.Tensor | None = None,
```
skyw: I think we always default to P2 norm for flattened. `grad_norm_square` would be a better name given grad_norm is so widely used.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -132,22 +139,37 @@ def _apply_moment2_normalization(
...
+        pre_orth_norm: torch.Tensor | None = None,
     ) -> torch.Tensor:
-        """Apply AdamW-style second moment accumulation and normalization.
+        """Apply second moment accumulation and normalization.
```
skyw, line 164: Good catch.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -132,22 +139,37 @@ def _apply_moment2_normalization(
         moment2: torch.Tensor,
         beta2: float,
         eps: float,
+        *,
```
skyw, line 160: Comment: Flag based arguments are never great. Fine to user internally to some extend. But if there are more flag dependent arguments are needed, should redesign the interface.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -175,6 +197,17 @@ def _apply_moment2_normalization(
             step_size = moment2.clamp_min(eps).rsqrt_()
             return orth_grad * step_size
 
+        elif self.moment2_method == "namo":
+            assert grad_fro_sq is not None and pre_orth_norm is not None, "NAMO requires grad_fro_sq and pre_orth_norm"
```
skyw: Fix: Raise RuntimeError here, don't assert.
Check style guide (google) for rules of assert.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -231,15 +269,26 @@ def step(self, closure: Callable[[], float] | None = None) -> float | None:
                     group_kwargs = {k: v for k, v in group.items() if k != "params"}
                     orth_grad = self.orthogonalize(p, grad, **group_kwargs)
 
-                update = self._apply_moment2_normalization(
-                    orth_grad=orth_grad,
-                    moment2=state["moment2_buffer"],
-                    beta2=group["beta2"],
-                    eps=group["eps"],
-                )
+                if self.moment2_method == "namo":
```
skyw: Q: why this dispatch is outside _apply_moment2_normalization? should be handled inside it if no particular reason.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -231,15 +269,26 @@ def step(self, closure: Callable[[], float] | None = None) -> float | None:
...
+                    # Capture ||pre_orth_grad||_F (after momentum + Nesterov)
+                    update = self._apply_moment2_normalization(
+                        orth_grad=orth_grad,
+                        moment2=state["moment2_buffer"],
+                        beta2=group["beta2"],
+                        eps=group["eps"],
+                        grad_fro_sq=grad_fro_sq,
+                        pre_orth_norm=torch.linalg.vector_norm(grad),
```
skyw: Q: grad_fro_sq and pre_orth_norm seems to be the same here?

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -231,15 +269,26 @@ def step(self, closure: Callable[[], float] | None = None) -> float | None:
...
+                    update = self._apply_moment2_normalization(
+                        orth_grad=orth_grad,
+                        moment2=state["moment2_buffer"],
+                        beta2=group["beta2"],
+                        eps=group["eps"],
+                    )
 
                 # perform weight update
-                # scale is applied to have update RMS == 1
+                # scale is applied to have update RMS roughly 1
```
skyw: It is not clear what scale means in this line. I'd just remove the comment including the line above.

tests/test_adaptive_muon.py
```diff
@@ -93,6 +94,9 @@ def test_second_moment_matches_shapes(self, shape, second_moment_method) -> None
             expected_shape = list(shape)
             expected_shape[avg_dim] = 1
             self.assertEqual(list(second_moment.shape), expected_shape)
+        elif second_moment_method == "namo":
+            # Scalar buffer
```
skyw: Fix: meaningless and confusing comment. Remove or explain more what it asserts.

skyw (changes requested):

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
mkhona-nvidia: Moment2MethodT type alias (L10-12): Extracted the Literal["adamuon", "normuon", "namo"] into a named type and exported it in __all__.

mkhona-nvidia: Class docstring updated (L17-27): Now mentions NAMO alongside AdaMuon/NorMuon, and removes the "AdamW-style or NorMuon-style" phrasing.

mkhona-nvidia: Signature uses type alias (L36): moment2_method: Moment2MethodT instead of inline Literal.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
mkhona-nvidia: NAMO scalar buffer (L44-46): torch.tensor(0.0, ...) instead of torch.zeros(()) for clarity.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
mkhona-nvidia: Renamed parameters (L54-57, L65-71): grad_fro_sq / pre_orth_norm became raw_grad / pre_orth_grad -- pass the raw tensors, not precomputed norms.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
mkhona-nvidia: assert replaced with raise RuntimeError (L79-81): Per Google style guide.

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
mkhona-nvidia: Norm computation moved inside _apply_moment2_normalization (L85-90): vector_norm(...).square() and vector_norm(...) are now internal to the method rather than computed at the call site.

mkhona-nvidia: Unified call site in step() (L99-138): Removed the if self.moment2_method == "namo" dispatch. Now there's a single _apply_moment2_normalization(...) call with raw_grad and pre_orth_grad passed through (they're None for non-NAMO methods).

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
mkhona-nvidia: Removed unclear comment (L129-130): "perform weight update / scale is applied to have update RMS roughly 1".

tests/test_adaptive_muon.py
mkhona-nvidia: done

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -121,6 +127,8 @@ def _initialize_moment2(
                 moment2_shape = list(grad.shape)
                 moment2_shape[avg_dim] = 1
                 moment2 = torch.zeros(moment2_shape, dtype=grad.dtype, device=grad.device)
+            elif self.moment2_method == "namo":
+                moment2 = torch.tensor(0.0, dtype=grad.dtype, device=grad.device)
```
skyw: In general, avoid empty shape tensors. see example below
```python
In [2]: a = torch.zeros(1)
In [3]: b = torch.tensor(0)

In [4]: b
Out[4]: tensor(0)
In [5]: a
Out[5]: tensor([0.])

In [6]: a.shape
Out[6]: torch.Size([1])
In [7]: b.shape
Out[7]: torch.Size([])
```
It is less error prone to compare size 1 instead of empty shape.

skyw (dismissed):
Recommend to remove empty tensor shape usage. otherwise LGTM

skyw (dismissed):

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -132,16 +138,16 @@ def _init_group(
...
                         raise ValueError(
                             f"{self.__class__.__name__} only supports 2D parameters, got shape {tuple(p.data.shape)}"
                         )
                     avg_dim = -1 if p.data.shape[-2] >= p.data.shape[-1] else -2
-                    # Specify the shape with reduced dimension
                     moment2_shape = list(p.data.shape)
                     moment2_shape[avg_dim] = 1
                     state["moment2_buffer"] = torch.zeros(moment2_shape, dtype=p.data.dtype, device=p.data.device)
+                elif self.moment2_method == "namo":
+                    state["moment2_buffer"] = torch.tensor(0.0, dtype=p.data.dtype, device=p.data.device)
```
skyw: What was he reason to rollback?

skyw (changes requested):

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
mkhona-nvidia: it isn't a rollback, it was an empty tensor and now it is a scalar initialized to 0

skyw: There was a version changed to use zeros per previous suggestion. test needs to be updated accordingly

skyw (changes requested):

emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
mkhona-nvidia: good point, let me revert

skyw (approved):

PR #165, opened 2026-04-24
ci: Fix container build on scheduled main branch run

chtruong814:
ci: Fix container build on scheduled main branch run

When `docs_only == True`, it would skip container build on CI runs on main branch, causing main branch CI runs to fail until a new commit that changes code is merged.

This change updates the Github Actions condition to be similar to what is used by other repos where the `needs.pre-flight.outputs.docs_only == 'false'` is removed on the container build job. On a PR when `docs_only == True`, then it will cause `cicd-wait-in-queue` to skip, it will also skip the container build.

### DIFF
--- a/.github/workflows/cicd-main.yml
+++ b/.github/workflows/cicd-main.yml
@@ -46,8 +46,9 @@ jobs:
     runs-on: ubuntu-latest
     environment: test
     if: |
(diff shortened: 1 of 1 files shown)


chtruong814:
Fast merging to resolve the CI issue

PR #170, opened 2026-04-29
Add batched Netwon Schulz

skyw:
use baddmm instead of addmm to handle (b, m, n) input.


FDecaYed (approved):
changes LGTM. Do we have benchmark result?

skyw:
> changes LGTM. Do we have benchmark result?

There is a benchmark script, we can add some.

PR #172, opened 2026-04-29
Add benchmark for Newton-Schulz

skyw:
Simple benchmark with TP support. TE path is untested as it requires building TE main, will wait for a release.
This is a benchmark only no functionality change.

Reorganized qwen config to be out of soap benchmark for future use.


skyw:
@greptile

PR #173, opened 2026-05-01
Relax threshold for flaky sinkhorn test

skyw:
The atol was too tight, it had roughly 10% chance of failure in CI.


mkhona-nvidia (dismissed):
how do you select these numbers?

skyw:
> how do you select these numbers?

Similar as others, tightest that can work for non essential tests.

mkhona-nvidia (approved):

PR #187, opened 2026-05-15
Fix NorMuon update norm scaling

mchrzanowski:
## Summary

  Fixes NorMuon update scaling in `AdaptiveMuon`.

  Previously, the `normuon` path applied row/column second-moment normalization to the raw orthogonalized update, then immediately applied the standard Muon scale factor. This can make the pre-scale
  adapted update have Frobenius norm near `sqrt(m*n)` instead of the raw orthogonalized update norm, so the final update can become larger than Muon by roughly `sqrt(max(m, n))`.

  This change re-matches the Frobenius norm of the Normuon-adapted update to the raw orthogonalized update before applying Muon scaling. That preserves the adaptive direction from Normuon while keeping
  the final update scale compatible with Muon.

  ## Changes

  - Added `_match_frobenius_norm(...)` helper.
  - Applied Frobenius norm re-matching in the `moment2_method="normuon"` path.
  - Added a regression test verifying that Normuon preserves the pre-Muon-scale Frobenius norm.

  ## Testing

  - `python -m py_compile emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py tests/test_adaptive_muon.py`

### DIFF
--- a/emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
+++ b/emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
@@ -123,8 +123,12 @@ class AdaptiveMuon(OrthogonalizedOptimizer):
         logging.debug(f"Applying Muon scale factor {scale_factor}, extra_scale_factor={self.extra_scale_factor}")
         return update * scale_factor * self.extra_scale_factor
 
-    @staticmethod
-    def _match_frobenius_norm(update: torch.Tensor, reference: torch.Tensor, eps: float) -> torch.Tensor:
+    def _match_frobenius_norm(
+        self,
+        update: torch.Tensor,
+        reference: torch.Tensor,
+        eps: float,
+    ) -> torch.Tensor:
         """Scale update to match the Frobenius norm of reference."""
         update_norm = torch.linalg.vector_norm(update)
         reference_norm = torch.linalg.vector_norm(reference)
--- a/emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
+++ b/emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
@@ -123,6 +123,13 @@ class AdaptiveMuon(OrthogonalizedOptimizer):
         logging.debug(f"Applying Muon scale factor {scale_factor}, extra_scale_factor={self.extra_scale_factor}")
         return update * scale_factor * self.extra_scale_factor
 
+    @staticmethod
+    def _match_frobenius_norm(update: torch.Tensor, reference: torch.Tensor, eps: float) -> torch.Tensor:
+        """Scale update to match the Frobenius norm of reference."""
+        update_norm = torch.linalg.vector_norm(update)
+        reference_norm = torch.linalg.vector_norm(reference)
+        return update * (reference_norm / update_norm.clamp_min(eps))
+
     @torch.no_grad()  # type: ignore[misc]
     @override
     def _init_group(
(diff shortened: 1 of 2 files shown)


emerging_optimizers/orthogonalized_optimizers/adaptive_muon.py
```diff
@@ -123,6 +123,13 @@ def _apply_muon_scale(self, update: torch.Tensor, size_out: int, size_in: int) -
         logging.debug(f"Applying Muon scale factor {scale_factor}, extra_scale_factor={self.extra_scale_factor}")
         return update * scale_factor * self.extra_scale_factor
 
+    @staticmethod
```
skyw: Q: why static?

If it is meant be called outside the class, it shouldn't be private with the leading "_" in function name.

tests/test_adaptive_muon.py
```diff
@@ -156,6 +156,35 @@ def test_moment2_accumulates_before_muon_update_scaling(self, moment2_method) ->
...
+        adaptive_opt = AdaptiveMuon(
+            [test_param],
+            lr=0.01,
+            momentum=0.0,
+            weight_decay=0.0,
+            moment2_method="normuon",
+            fp32_matmul_prec="highest",
+        )
+
+        update = adaptive_opt._apply_moment2_normalization(
```
skyw: Need to call _init_group or first step. testing _apply_moment2_normalization alone is not enough to guarantee correct initialization. 
Fine to keep unit test for _apply_moment2_normalization. but need to test the lazy init.

tests/test_adaptive_muon.py
```diff
@@ -156,6 +156,35 @@ def test_moment2_accumulates_before_muon_update_scaling(self, moment2_method) ->
...
+            orth_grad=orth_grad,
+            moment2=moment2,
+            beta2=0.0,
+            eps=1e-8,
+        )
+
+        torch.testing.assert_close(
+            torch.linalg.vector_norm(update),
+            torch.linalg.vector_norm(orth_grad),
+            atol=1e-6,
```
skyw: Should this match exact? use 0 for atol, rtol if so.

skyw (changes requested):

tests/test_adaptive_muon.py
```diff
@@ -156,6 +156,35 @@ def test_moment2_accumulates_before_muon_update_scaling(self, moment2_method) ->
...
+            rtol=1e-6,
+        )
```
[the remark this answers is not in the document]
skyw: Valid concern from AI. Test is not sufficient @mchrzanowski

skyw:
LGTM.
Defer to @mkhona-nvidia to take a look at the algorithm and approve.

tests/test_adaptive_muon.py
```diff
@@ -156,6 +156,59 @@ def test_moment2_accumulates_before_muon_update_scaling(self, moment2_method) ->
             rtol=1e-6,
         )
 
+    @parameterized.parameters({"shape": (16, 4)}, {"shape": (4, 16)})
+    def test_normuon_step_preserves_pre_scale_frobenius_norm(self, shape) -> None:
+        """Test that NorMuon preserves the raw orthogonalized update scale through step()."""
+
+        class CaptureUpdateAdaptiveMuon(AdaptiveMuon):
```
skyw: It is a good idea to preserve input for exact match.
But this is a "hack" that changes intended use of member functions and should not be used. They will also fail if we type check tests.

Value can be restored if every step is exactly representable. For example, a * lr / lr may not get exact a back if lr =0.01. but it will if a doesn't have a lot of mantissa bits populated and lr is power of 2, 0.125 for example.
```python
In [2]: 3.1415 * 0.01 / 0.01
Out[2]: 3.1415000000000006

In [3]: 3.1415 * 0.0125 / 0.0125
Out[3]: 3.1415

```

skyw (changes requested):

skyw (approved):

mkhona-nvidia (approved):

PR #206, opened 2026-05-22
LAprop as an optimizer along with frobenius weight normalization + associated tests

mkhona-nvidia:
Convert the LAProp update function into a scalar optimizer

### DIFF
--- a/emerging_optimizers/scalar_optimizers/laprop.py
+++ b/emerging_optimizers/scalar_optimizers/laprop.py
@@ -167,6 +167,7 @@ class LaProp(WeightDecayMixin, torch.optim.Optimizer):
                 grad = p.grad
                 state = self.state[p]
                 state["step"] += 1
+                pre_norm = p.data.norm() if self.normalize else None
 
                 self._apply_weight_decay_inplace(p.data, grad, lr, weight_decay)
 
@@ -179,7 +180,6 @@ class LaProp(WeightDecayMixin, torch.optim.Optimizer):
                     state["step"],
                     eps,
                 )
-                pre_norm = p.data.norm() if self.normalize else None
                 p.data.add_(update, alpha=-lr)
                 if pre_norm is not None:
                     p.data.mul_(pre_norm / p.data.norm().clamp_min(self.normalize_eps))
--- a/emerging_optimizers/scalar_optimizers/laprop.py
+++ b/emerging_optimizers/scalar_optimizers/laprop.py
@@ -65,8 +65,7 @@ class LaProp(WeightDecayMixin, torch.optim.Optimizer):
         eps: Term added to the denominator for numerical stability.
         weight_decay: Weight decay coefficient.
         correct_bias: Whether to apply bias correction to the first and second moment EMAs.
-        normalize: Whether to normalize each updated parameter back to its pre-update Frobenius norm.
-        normalize_eps: Term used to avoid division by zero during parameter normalization.
+        frob_normalize: Whether to normalize each updated parameter back to its pre-update Frobenius norm.
         weight_decay_method: Method to apply weight decay, see
             :class:`~emerging_optimizers.mixin.WeightDecayMixin` for more details.
     """
@@ -79,8 +78,7 @@ class LaProp(WeightDecayMixin, torch.optim.Optimizer):
         eps: float = 1e-8,
         weight_decay: float = 0.0,
         correct_bias: bool = True,
-        normalize: bool = False,
-        normalize_eps: float = 1e-8,
+        frob_normalize: bool = False,
         weight_decay_method: WeightDecayT = "decoupled",
     ) -> None:
         if not 0.0 <= lr:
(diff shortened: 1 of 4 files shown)


skyw:
Shall we seek more programmable way to make optimizer out of scalar update function? Instead of adding them one by one (with a lot of duplicated code) when we need it?

emerging_optimizers/scalar_optimizers/laprop.py
```diff
@@ -0,0 +1,187 @@
...
+    def __init__(
+        self,
+        params: ParamsT,
+        lr: float = 1e-3,
+        betas: tuple[float, float] = (0.9, 0.999),
+        eps: float = 1e-8,
+        weight_decay: float = 0.0,
+        correct_bias: bool = True,
+        normalize: bool = False,
+        normalize_eps: float = 1e-8,
```
skyw: Nit: Do we need two eps? They are for different things, but I feel having two of them only increase probability to mess up one of them.

emerging_optimizers/scalar_optimizers/laprop.py
```diff
@@ -0,0 +1,187 @@
...
+
+        Args:
+            closure: A closure that reevaluates the model and returns the loss (optional).
+
+        Returns:
+            The loss from the closure, if provided.
+        """
+        loss = None
+        if closure is not None:
+            with torch.enable_grad():
```
skyw: I know greptile like to signal this, but all of our optimizer are not supposed to have gradient in it. can just assert closure is None and ignore greptile.

emerging_optimizers/scalar_optimizers/laprop.py
```diff
@@ -0,0 +1,187 @@
...
+            weight_decay = group["weight_decay"]
+            correct_bias = group["correct_bias"]
+
+            for p in group["params"]:
+                if p.grad is None:
+                    continue  # pragma: no cover
+
+                grad = p.grad
+                state = self.state[p]
+                state["step"] += 1
```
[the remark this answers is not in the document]
skyw: Seems a valid concern?
I'd say either fix or just remove weight_decay_method in argument and only use decoupled.

emerging_optimizers/scalar_optimizers/laprop.py
```diff
@@ -0,0 +1,187 @@
...
+                    state["exp_avg"],
+                    state["exp_avg_sq"],
+                    correct_bias,
+                    betas,
+                    state["step"],
+                    eps,
+                )
+                pre_norm = p.data.norm() if self.normalize else None
+                p.data.add_(update, alpha=-lr)
+                if pre_norm is not None:
```
skyw: Use self.normalize as condition.

emerging_optimizers/scalar_optimizers/laprop.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def __init__(
+        self,
+        params: ParamsT,
+        lr: float = 1e-3,
+        betas: tuple[float, float] = (0.9, 0.999),
+        eps: float = 1e-8,
+        weight_decay: float = 0.0,
+        correct_bias: bool = True,
+        normalize: bool = False,
```
skyw: Nit: Consider a more informative name to indicate what normalization the flag enables.

tests/test_laprop.py
```diff
@@ -0,0 +1,185 @@
...
+        optimizer.step()
+
+        self.assertEqual(optimizer.state[param]["step"], 1)
+        self.assertIn("exp_avg", optimizer.state[param])
+        self.assertIn("exp_avg_sq", optimizer.state[param])
+
+        expected_exp_avg_sq = (1 - beta2) * grad.square()
+        normalized_grad = grad / (grad.abs() + optimizer.param_groups[0]["eps"])
+        expected_exp_avg = (1 - beta1) * normalized_grad
+        torch.testing.assert_close(optimizer.state[param]["exp_avg_sq"], expected_exp_avg_sq)
```
skyw: Q: Should they exact match?

tests/test_laprop.py
```diff
@@ -0,0 +1,185 @@
...
+        old_param = param.detach().clone()
+        exp_avg = torch.zeros_like(param)
+        exp_avg_sq = torch.zeros_like(param)
+        expected_update = calculate_laprop_update(grad, exp_avg, exp_avg_sq, True, betas, 1, eps)
+
+        param.grad = grad.clone()
+        optimizer.step()
+
+        torch.testing.assert_close(param, old_param - lr * expected_update)
+        torch.testing.assert_close(optimizer.state[param]["exp_avg"], exp_avg)
```
skyw: Test name suggest "match", both atol and rtol needs to be 0.

skyw (changes requested):

skyw (dismissed):
We should do a full repo clean up for the closure. We don't support it in any of our optimizer, i.e. none was tested, should just error out if closure is not none.

skyw (approved):

skyw (approved):

PR #217, opened 2026-06-02
Add Muon Cubic5 coefficients

huanghua1994:
This PR adds a relaxed cubic Newton-Schulz coefficients for Muon. Please refer to https://arxiv.org/abs/2606.00371 for the detailed information, including the construction of the coefficients and model training test results.

### DIFF
--- a/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
+++ b/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
@@ -207,14 +207,11 @@ def newton_schulz(
     else:
         raise ValueError(f"Invalid coefficient type: {coefficient_type}")
 
-    if coefficient_type == "cubic5" and steps > len(coefficient_sets):
-        logging.warning(
-            "cubic5 is a fixed %d-step schedule, but steps=%d was requested. "
-            "Skipping the extra Newton-Schulz steps.",
-            len(coefficient_sets),
-            steps,
+    if coefficient_type == "cubic5" and steps != len(coefficient_sets):
+        raise ValueError(
+            f"cubic5 is a fixed {len(coefficient_sets)}-step schedule; got steps={steps}. "
+            "Use steps=5, or pass explicit custom coefficients."
         )
-        steps = len(coefficient_sets)
 
     repeat_last_types = ("polar_express", "cans", "deepseekv4")
     iter_mode: CoeffIterMode = "repeat_last" if coefficient_type in repeat_last_types else "cycle"
--- a/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
+++ b/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
@@ -25,7 +25,7 @@ __all__ = ["newton_schulz", "newton_schulz_tp", "NSCoeffT", "get_coefficient_ite
 
 CoeffIterMode = Literal["cycle", "repeat_last"]
 
-NSCoeffT = Literal["simple", "quintic", "polar_express", "cans", "aol", "deepseekv4", "custom"]
+NSCoeffT = Literal["simple", "quintic", "cubic5", "polar_express", "cans", "aol", "deepseekv4", "custom"]
 
 _COEFFICIENT_SETS = {
     # Values are rounded to closest representable in single precision.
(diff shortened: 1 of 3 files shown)


emerging_optimizers/orthogonalized_optimizers/muon_utils.py
```diff
@@ -148,6 +160,7 @@ def newton_schulz(
     Parameter ``coefficient_type`` can be one of the following
       - "simple": Default coefficient set.
       - "quintic": Quintic iteration with optimized coefficients.
+      - "cubic5": Relaxed cubic five-step schedule.
```
skyw: Add to the last, although not explicit, we've been using first come first serve when adding new type.

emerging_optimizers/orthogonalized_optimizers/muon_utils.py
```diff
@@ -194,7 +207,7 @@ def newton_schulz(
     else:
         raise ValueError(f"Invalid coefficient type: {coefficient_type}")
 
-    repeat_last_types = ("polar_express", "cans", "deepseekv4")
+    repeat_last_types = ("cubic5", "polar_express", "cans", "deepseekv4")
```
skyw: Also append to the right most.

emerging_optimizers/orthogonalized_optimizers/muon_utils.py
```diff
@@ -315,6 +328,8 @@ def newton_schulz_step(
     A = X @ X.mT
     if tp_group is not None:
         torch.distributed.all_reduce(A, op=torch.distributed.ReduceOp.SUM, group=tp_group)
+    if c == 0.0:
+        return torch.addmm(X, A, X, alpha=b, beta=a)
```
skyw: Nit: I prefer
```python
if c != 0.0:
  B = torch.baddbmm(A, A, A, alpha=c, beta=b)
  X = torch.baddbmm(X, B, X, alpha=1.0, beta=a)
else:
  X = torch.baddbmm(X, A, X, alpha=b, beta=a)
```

skyw (changes requested):
Code looks ok except a few minor suggestions.

Need same level of testing as existing paths.

Future note, I don't like using c == 0 as flag to dispatch to different path. There are so many small things we've added over time. Could use a redesign at some point.

emerging_optimizers/orthogonalized_optimizers/muon_utils.py
huanghua1994: Fixed as suggested.

emerging_optimizers/orthogonalized_optimizers/muon_utils.py
huanghua1994: Fixed as suggested.

emerging_optimizers/orthogonalized_optimizers/muon_utils.py
huanghua1994: Fixed as suggested.

emerging_optimizers/orthogonalized_optimizers/muon_utils.py
```diff
@@ -194,6 +207,15 @@ def newton_schulz(
     else:
         raise ValueError(f"Invalid coefficient type: {coefficient_type}")
 
+    if coefficient_type == "cubic5" and steps > len(coefficient_sets):
+        logging.warning(
```
skyw: Implicitly changing argument is dangerous, no one reads warnings. If it is only supposed to be used for 5 steps, raise value error if step is not 5.

tests/test_muon_utils.py
```diff
@@ -133,6 +133,63 @@ def test_newtonschulz_3d_input_close_to_per_slice(self, batch, dim1, dim2):
...
+            out_cubic5_ref,
+            atol=1e-6,
+            rtol=1e-7,
+        )
+
+    @parameterized.product(
+        size=[(511, 513), (511, 257), (257, 513)],
+        steps=[1, 2, 3, 4],
+    )
+    def test_cubic5_prefix_close_to_reference(self, size, steps):
```
skyw: This can be removed after switching to raise error for steps != 5. Need one test to cover right error is raised.

tests/test_muon_utils.py
```diff
@@ -133,6 +133,63 @@ def test_newtonschulz_3d_input_close_to_per_slice(self, batch, dim1, dim2):
...
+        torch.testing.assert_close(
+            out_cubic5_test,
+            out_cubic5_ref,
+            atol=1e-6,
+            rtol=1e-7,
+        )
+
+    def test_cubic5_too_many_steps_matches_five_step_schedule(self) -> None:
+        x = torch.randn(5, 7, device=self.device, dtype=torch.float32)
+        out_cubic5_6 = muon_utils.newton_schulz(x, steps=6, coefficient_type="cubic5")
```
skyw: Same as above. Just needs one test to ensure raised messaged for wrong number of steps.

tests/test_muon_utils.py
```diff
@@ -368,6 +425,22 @@ def test_batched_newton_schulz_step_close_to_unbatched(self, batch, dim1, dim2):
 
         torch.testing.assert_close(batched, per_item, atol=1e-8, rtol=0)
 
+    @parameterized.parameters(
+        (2, 16, 16),
+        (4, 16, 32),
+        (3, 32, 16),
+    )
+    def test_batched_cubic_newton_schulz_step_matches_formula(self, batch, dim1, dim2):
```
skyw: `matches` in test name means exact match, i.e. atol=0, rtol=0. If they can't match, use `_closes_to` in test name.

skyw (changes requested):

tests/test_muon_utils.py
huanghua1994: Renamed to "_close_to". It still has very small numerical differences due to calculation order. Test passed.

tests/test_muon_utils.py
huanghua1994: Fixed as suggested.

emerging_optimizers/orthogonalized_optimizers/muon_utils.py
huanghua1994: Fixed as suggested.

skyw (approved):

PR #224, opened 2026-06-04
Refactor of pre/post weight update hook functions

mkhona-nvidia:
# Weight Update Hooks

This change adds a small `weight_update_hooks` library for reusable behavior around an optimizer's final in-place parameter update. A hook is a configured object passed into an optimizer, for example:

```python
from emerging_optimizers.weight_update_hooks import RadialBrake

optimizer = Muon(
    params,
    weight_decay=0.0,
    weight_update_hook=RadialBrake(outward_scale_factor=0.5),
)
```

The optimizer owns only one hook object. Hook-specific arguments live on the hook constructor, not on the optimizer constructor. The base update flow is:

```python
pre_update_state = weight_update_hook.pre_weight_update_inplace(p, update)
p.add_(update, alpha=-lr)
post_weight_update_fn_inplace(p)
weight_update_hook.post_weight_update_inplace(p, pre_update_state)
```

`pre_update_state` is a transient value returned by the pre hook and immediately consumed by the post hook for the same parameter. It is not stored in `optimizer.state`, so it does not pollute checkpoints.

The library currently provides three hook implementations, including the newly added `RadialBrake` based on https://nilin.github.io/radial-brake/:

- `NoOpWeightUpdateHook`: default no-op behavior.
- `Hyperball`: normalizes the update to a target radius before the update, then projects the updated weight back to that radius.
- `RadialBrake`: applies the normal optimizer update, then rescales the updated weight so radial norm changes are damped. For pre-update weight `w_prev` and updated weight `w`, it sets:

```math
\|w_{\text{brake}}\| = \|w_{\text{prev}}\| + s(\|w\| - \|w_{\text{prev}}\|)
```

where `s = outward_scale_factor` if the update increases the norm, otherwise `s = inward_scale_factor`.

`MuonHyperball` now uses this shared hook machinery by passing `Hyperball(radius=hyperball_radius, eps=hyperball_eps)` into `Muon`, instead of implementing custom pre/post update methods and storing temporary values in optimizer state.


emerging_optimizers/orthogonalized_optimizers/orthogonalized_optimizer.py
```diff
@@ -195,8 +199,10 @@ def step(self, closure: Callable[[], float] | None = None) -> float | None:
 
                 # perform weight update with pre and post weight update functions for subclass customization
                 self.pre_weight_update_fn_inplace(p, orth_grad)
+                weight_update_hook_pre_update_state = self.weight_update_hook.pre_weight_update_inplace(p, orth_grad)
```
skyw: name is too long, two update in the name.

emerging_optimizers/weight_update_hooks/base.py
```diff
@@ -0,0 +1,56 @@
...
+# limitations under the License.
+from typing import Protocol
+
+import torch
+
+
+__all__ = ["NoOpWeightUpdateHook", "WeightUpdateHook"]
+
+
+class WeightUpdateHook(Protocol):
```
skyw: @mkhona-nvidia I don't oppose using Protocol. But I think need to get yourself comfortable with PEP544 before this can be merged.

emerging_optimizers/weight_update_hooks/hyperball.py
```diff
@@ -0,0 +1,63 @@
...
+    ) -> None:
+        self.radius = radius
+        self.eps = eps
+
+    def pre_weight_update_inplace(
+        self,
+        p: torch.Tensor,
+        update: torch.Tensor,
+    ) -> torch.Tensor:
+        current_norm = torch.linalg.vector_norm(p.detach().to(torch.float32))
```
skyw: detach is not necessary as all of our optimizers are wrapped in no_grad.

emerging_optimizers/weight_update_hooks/hyperball.py
```diff
@@ -0,0 +1,63 @@
...
+
+        if self.radius is not None:
+            radius = torch.as_tensor(self.radius, device=p.device, dtype=torch.float32)
+        else:
+            if current_norm.item() == 0:
```
skyw: This triggers synced device to host copy. Can't have it in every call. We may have it already, but it needs to be fixed. Otherwise can't use in any pratical runs.

emerging_optimizers/weight_update_hooks/hyperball.py
```diff
@@ -0,0 +1,63 @@
...
+    """Normalize update and post-update weights to a fixed Frobenius norm.
+
+    This hook mirrors the hyperball-style behavior used by MuonHyperball: before the weight update, normalize the
+    update to the target radius; after the weight update, project the parameter back to that radius.
+    """
+
+    def __init__(
+        self,
+        radius: float | None = None,
+        eps: float = 1e-8,
```
skyw: Maybe reduce this to 1e-15?

emerging_optimizers/weight_update_hooks/hyperball.py
```diff
@@ -0,0 +1,63 @@
...
+        update_norm = torch.linalg.vector_norm(update.to(torch.float32)).clamp_min(self.eps)
+        update.mul_((radius / update_norm).to(dtype=update.dtype))
+        return radius
+
+    def post_weight_update_inplace(
+        self,
+        p: torch.Tensor,
+        pre_update_state: torch.Tensor | None,
+    ) -> None:
+        if pre_update_state is None:
```
skyw: Q: why does it allow None and only to raise error? making it non optional is not sufficient?

emerging_optimizers/weight_update_hooks/radial_brake.py
```diff
@@ -0,0 +1,67 @@
...
+
+        \\|w_{brake}\\| = \\|w_{prev}\\| + s(\\|w\\| - \\|w_{prev}\\|)
+
+    where ``s`` is ``outward_scale_factor`` when the update increases the norm, otherwise
+    ``inward_scale_factor``.
+    """
+
+    def __init__(
+        self,
+        outward_scale_factor: float = 0.5,
```
skyw: scale_factor looks redundant, would outward_scale be ambiguous? if not, use the short form

emerging_optimizers/weight_update_hooks/radial_brake.py
```diff
@@ -0,0 +1,67 @@
...
+        self.outward_scale_factor = outward_scale_factor
+        self.inward_scale_factor = inward_scale_factor
+        self.eps = eps
+
+    def pre_weight_update_inplace(
+        self,
+        p: torch.Tensor,
+        update: torch.Tensor,
+    ) -> torch.Tensor:
+        return torch.linalg.vector_norm(p.detach().to(torch.float32))
```
skyw: why eps is not used here?

emerging_optimizers/weight_update_hooks/radial_brake.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def post_weight_update_inplace(
+        self,
+        p: torch.Tensor,
+        pre_update_state: torch.Tensor | None,
+    ) -> None:
+        if pre_update_state is None:
```
skyw: Same question as before, what's the reason to allow None and only to raise error?

emerging_optimizers/weight_update_hooks/radial_brake.py
```diff
@@ -0,0 +1,67 @@
...
+            raise RuntimeError("RadialBrake requires pre-update norm state")
+        pre_norm = pre_update_state
+        post_norm = torch.linalg.vector_norm(p.detach().to(torch.float32))
```
skyw: Don't detach.

emerging_optimizers/weight_update_hooks/radial_brake.py
```diff
@@ -0,0 +1,67 @@
...
+        norm_delta = post_norm - pre_norm
+        scale_factor = self.outward_scale_factor if norm_delta.item() > 0 else self.inward_scale_factor
```
skyw: same as before, don't use .item().

tests/test_orthogonalized_optimizer.py
```diff
@@ -434,6 +434,17 @@ def test_zero_norm_raises_error(self) -> None:
         with self.assertRaises(ValueError):
             muon_hyperball.MuonHyperball([test_param], lr=0.01)
 
+    def test_rejects_weight_update_hook(self) -> None:
+        """MuonHyperball manages its own Hyperball hook internally."""
+        test_param = nn.Parameter(torch.randn((5, 7), dtype=torch.float32, device=self.device))
+
+        with self.assertRaisesRegex(TypeError, "does not accept a 'weight_update_hook' argument"):
```
skyw: This is better to be a KeyError,

tests/test_weight_update_hooks.py
```diff
@@ -0,0 +1,135 @@
...
+        hook = NoOpWeightUpdateHook()
+        param = torch.tensor([3.0, 4.0], device=self.device)
+        update = torch.tensor([1.0, -2.0], device=self.device)
+        param_before = param.clone()
+        update_before = update.clone()
+
+        pre_update_state = hook.pre_weight_update_inplace(param, update)
+        hook.post_weight_update_inplace(param, pre_update_state)
+
+        torch.testing.assert_close(param, param_before, atol=0.0, rtol=0.0)
```
skyw: Good to test exact match.

We should probably have a partial function called assert_equal given it is so widely used.

tests/test_weight_update_hooks.py
```diff
@@ -0,0 +1,135 @@
...
+        torch.testing.assert_close(update, update_before, atol=0.0, rtol=0.0)
+
+    def test_radial_brake_dampens_outward_norm_change(self) -> None:
```
skyw: Suggestion: always state expected behavior in test name. e.g. in this case, what close to what. "change" is too vague of a behavior.

emerging_optimizers/orthogonalized_optimizers/orthogonalized_optimizer.py
```diff
@@ -103,6 +105,7 @@ def __init__(
         weight_decay_method: opt_mixin.WeightDecayT,
         fp32_matmul_prec: FP32MatmulPrecT,
         scaled_orthogonalize_fn: Callable | None = None,
+        weight_update_hook: WeightUpdateHook | None = None,
```
skyw: Let's make it private, ```_weight_update_hook```, to suggest it is not supposed to be modified after initialization.

Hook easily invites abuse. making it private at least making people aware of the abusing. A setter can be provided if we want to support properly change the hook inflight.

skyw (changes requested):
Code logic LGTM. A lot of test names can be improved, not super critical.

Before merge, please get familiar with protocol, i.e. being able to explain what purpose it serves here and why not use subclass.

One critical thing is 0 handling, i.e. epsilon.
* at least, it should be consistently applied, for vector_norm for example.
* Comparing a floating point number directly against zero (a == 0 for example) is actually testing underflow, not numerical 0. Same eps apply, if numbers smaller than eps are considered 0, logical test to determine a numerical 0 should be abs(a) < eps
* Magnitude aware handling maybe out of scope, but something to consider as we running into this more and more.

emerging_optimizers/weight_update_hooks/hyperball.py
skyw: Actually, vector_norm has an dtype argument, explicit to fp32 is not necessary.

It technically not dtype but compute type though, don't know who added it to pytorch.

emerging_optimizers/weight_update_hooks/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from emerging_optimizers.weight_update_hooks.base import NoOpWeightUpdateHook, WeightUpdateHook
+from emerging_optimizers.weight_update_hooks.hyperball import Hyperball
+from emerging_optimizers.weight_update_hooks.radial_brake import RadialBrake
+from emerging_optimizers.weight_update_hooks.relative_update import RelativeUpdate
+
+
+__all__ = [
+    "Hyperball",
```
skyw: Q: why some has Hook while others don't? 
If there is no particular reason, make naming consistent.

emerging_optimizers/weight_update_hooks/base.py
```diff
@@ -0,0 +1,65 @@
...
+__all__ = ["NoOpWeightUpdateHook", "WeightUpdateHook"]
+
+
+HookStateT = TypeVar("HookStateT")
+
+
+class WeightUpdateHook(Protocol[HookStateT]):
+    """Static structural contract for behavior around the final weight update.
+
+    PEP 544 structural typing lets third-party hooks satisfy this interface without
```
skyw: too verbose and not always accurate, remove.

emerging_optimizers/orthogonalized_optimizers/muon.py
```diff
@@ -86,6 +87,7 @@ def __init__(
         scale_mode: MuonScaleT = "spectral",
         extra_scale_factor: float = 1.0,
         use_syrk: bool = False,
+        weight_update_hook: WeightUpdateHook[Any] | None = None,
```
skyw: Q: what's the reason to use Any while a HookStateT is defined?
can set bound to HookStateT to accept known types it needs to handle.

emerging_optimizers/weight_update_hooks/radial_brake.py
```diff
@@ -0,0 +1,84 @@
...
+import torch
+
+
+__all__ = ["RadialBrake"]
+
+
+class RadialBrake:
+    """Dampen radial norm changes after an optimizer update.
+
+    The optimizer first applies its usual update ``w = w_prev + dw``. This hook then rescales ``w`` so that
```
skyw, line 26: Add argument section.

emerging_optimizers/weight_update_hooks/relative_update.py
```diff
@@ -0,0 +1,62 @@
...
+    .. math::
+
+        U \leftarrow \frac{\lVert W\rVert_F}{\lVert U\rVert_F} U.
+
+    Therefore the learning rate directly controls the relative update norm. Unlike
+    :class:`Hyperball`, this hook does not project the weight after the update.
+    """
+
+    def __init__(self, eps: float = 1e-15) -> None:
+        if not math.isfinite(eps) or eps <= 0.0:
```
skyw: This is overkill, check magnitude on eps would be enough.
It would be strange to check if values are finite at one place while not everywhere else

tests/test_weight_update_hooks.py
```diff
@@ -0,0 +1,234 @@
...
+    def test_radial_brake_halves_outward_norm_increase(self) -> None:
+        hook = RadialBrake(outward_scale=0.5, inward_scale=1.0)
+        param = torch.tensor([3.0, 4.0], device=self.device)
+        update = torch.tensor([3.0, 4.0], device=self.device)
+
+        pre_update_state = hook.pre_weight_update_inplace(param, update)
+        param.add_(update)
+        hook.post_weight_update_inplace(param, pre_update_state)
+
+        torch.testing.assert_close(torch.linalg.vector_norm(param), torch.tensor(7.5, device=self.device))
```
skyw: Comment why it should be 7.5

tests/test_weight_update_hooks.py
```diff
@@ -0,0 +1,234 @@
...
+    def test_radial_brake_retains_twenty_percent_of_inward_norm_decrease(self) -> None:
+        hook = RadialBrake(outward_scale=1.0, inward_scale=0.2)
+        param = torch.tensor([6.0, 8.0], device=self.device)
+        update = torch.tensor([-3.0, -4.0], device=self.device)
+
+        pre_update_state = hook.pre_weight_update_inplace(param, update)
+        param.add_(update)
+        hook.post_weight_update_inplace(param, pre_update_state)
+
+        torch.testing.assert_close(torch.linalg.vector_norm(param), torch.tensor(9.0, device=self.device))
```
skyw: Same as above. always comment hardcoded test target.

tests/test_weight_update_hooks.py
```diff
@@ -0,0 +1,234 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
skyw, line 1: Too many hardcoded simple tests don't do a lot different, please trim.

Also think about how to improve tests coverage, includes but not limits to:
* Coverage, a lot of tests only test shape 2 input.
* Could it be exact match? can manipulate to value to have exactly representable norm for example.

skyw (changes requested):
Bottom line is it can be merged after trimming the tests.

emerging_optimizers/weight_update_hooks/hyperball.py
```diff
@@ -0,0 +1,62 @@
...
+            raise ValueError(f"radius must be finite and at least eps={eps}, got {radius}")
+        self.radius = radius
+        self.eps = eps
+
+    def _scale_to_radius_inplace(self, tensor: torch.Tensor) -> None:
+        norm = torch.linalg.vector_norm(tensor, dtype=torch.float32)
+        is_numerical_zero = norm < self.eps
+        radius = torch.as_tensor(self.radius, device=tensor.device, dtype=torch.float32)
+        scale = torch.where(is_numerical_zero, torch.zeros_like(norm), radius / norm.clamp_min(self.eps))
+        tensor.mul_(scale.to(dtype=tensor.dtype))
```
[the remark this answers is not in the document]
skyw: @mkhona-nvidia maybe comment it is intentional to mask numerical 0 to true 0 here.

skyw (approved):
Tests still need some rework. let's merge this one first, then do another to improve tests.

PR #230, opened 2026-06-17
fix(muon_utils): keep newton_schulz scale-invariant for small-norm inputs (#229)

### LINKED ISSUE #229: newton_schulz silently degenerates (non-orthogonal output) for small-norm inputs due to F.normalize eps floor
**Describe the bug**

`newton_schulz` is expected to be scale-invariant: for any input scale it should return an (semi-)orthogonal matrix with `‖UVᵀ‖_F ≈ √min(m,n)`. It does not. The prelude `X = F.normalize(x, p=2, dim=(-2,-1), eps=1e-7)` divides by `eps` instead of `‖x‖_F` once `‖x‖_F < eps`, so the normalized matrix has norm `‖x‖_F/eps ≪ 1`. The Newton–Schulz iteration (tuned for singular values ≈ 1) cannot lift it, and the output is a **degenerate, non-orthogonal matrix whose norm collapses toward 0** — silently, with no warning or error.

**Steps/Code to reproduce bug**

Minimal, torch-only (no GPU, runs in seconds) — mirrors the `F.normalize(..., eps=1e-7)` prelude exactly:

```python
import torch, torch.nn.functional as F
torch.manual_seed(0); torch.set_float32_matmul_precision("high")

def newton_schulz(x, steps=6, eps=1e-7):
    a, b, c = 3.4445, -4.7750, 2.0315
    X = F.normalize(x, p=2, dim=(-2, -1), eps=eps)   # the line under scrutiny
    for _ in range(steps):
        A = X @ X.transpose(-2, -1)
        X = a * X + (b * A + c * (A @ A)) @ X
    return X

m = 2048; base = torch.randn(m, m); base /= base.norm()   # unit-Frobenius direction
for N in (1e0, 1e-6, 1e-8, 1e-9, 1e-10):
    print(f"||in||_F={N:>7.0e}  ||out||_F={newton_schulz(base*N).norm():8.3f}  (ideal ~{m**0.5:.1f})")
```

Output:
```
||in||_F=  1e+00  ||out||_F=  39.929   (ideal ~45.3)
||in||_F=  1e-06  ||out||_F=  39.929
||in||_F=  1e-08  ||out||_F=  40.118
||in||_F=  1e-09  ||out||_F=  16.135   <- DEGENERATE
||in||_F=  1e-10  ||out||_F=   1.669   <- DEGENERATE
```

Same against the package itself (`from emerging_optimizers.orthogonalized_optimizers.muon_utils import newton_schulz`, `coefficient_type="polar_express"`): `1e0→44.93, 1e-8→42.11, 1e-10→1.83`.

**Expected behavior**

A scale-invariant orthogonalizer should return `‖out‖_F ≈ √min(m,n)` regardless of input scale (the plateau value), instead of collapsing once `‖x‖_F` underflows the eps floor. At minimum, it should warn/error rather than silently emit a non-orthogonal result.

**Additional context**

- Location: `emerging_optimizers/orthogonalized_optimizers/muon_utils.py`, `newton_schulz()`, `eps: float = 1e-7` used in `F.normalize`.
- Suggested fix: use `eps=1e-30` purely as a divide-by-zero guard, or normalize by `‖x‖_F.clamp_min(tiny)` with `tiny` near the dtype floor, or detect/warn when `‖x‖_F < eps`.
- Why it matters: combined with a framework that applies gradient-norm clipping to the Muon param group (e.g. Megatron-LM `ChainedOptimizer`), the clip coefficient scales per-matrix gradients below this floor → Newton–Schulz silently emits degenerate updates → training stalls with the forward/loss looking completely normal (very hard to diagnose). Filed separately on Megatron-LM; the two interact.
- Environment: Emerging-Optimizers 0.4.0a0, PyTorch 2.11+cu128 (repro above: any torch, CPU).

yuchenwang3:
## What

`newton_schulz` is meant to be scale-invariant — the orthogonalization of `x` and `c*x` should be identical. It is not: the internal `F.normalize(x, p=2, dim=(-2, -1), eps=1e-7)` divides a small-norm input by `eps` instead of its norm once `||x||_F < eps`, so `||X||_F = ||x||_F / eps << 1` and the Newton–Schulz iteration (tuned for singular values ~1) cannot lift it → a **silently** degenerate, non-orthogonal output (no warning, no error).

## Fix

Lower the `eps` default `1e-7 → 1e-30` so it acts purely as a divide-by-zero guard, which is its documented purpose (`eps: Small constant to avoid division by zero.`). Input is enforced `float32`, so `1e-30` is a safe guard. This covers all paths, since they share this `eps`:
- non-TP: `torch.nn.functional.normalize(..., eps=eps)`
- TP: `distributed_normalize_p2(x, eps, tp_group)` (`x / sqrt(sum).clamp_min(eps)`)
- `newton_schulz_tp` routes through `newton_schulz`, so it inherits the fix.

## Before / after (standalone, CPU, mirrors the `F.normalize(..., eps=1e-7)` prelude)

```
eps=1e-7 :  ||in||_F=1e-9 -> ||out||_F=16.1 ,  1e-10 -> 1.67    (ideal ~45)   <- DEGENERATE
eps=1e-30:  ||out||_F stays ~45 across all input scales                       <- scale-invariant
```

## Test

Adds `test_newtonschulz_scale_invariance` (asserts the orthogonalized output is identical for `x` and `scale * x`, `scale ∈ {1e-2, 1e-6, 1e-9, 1e-12}`). It fails on the old `eps=1e-7` default and passes with the fix.

## Alternatives (happy to switch to maintainers' preference)

1. `eps = torch.finfo(x.dtype).tiny` (dtype-relative guard).
2. Warn / raise when `||x||_F < eps` instead of changing the default (keeps current behavior but stops the silent failure).

## Note

Could not run the test suite locally (macOS arm64 has no `triton` wheel, so the package fails to import); relying on CI. The standalone repro above was verified on CPU.

## Why it matters

Combined with a training framework that applies gradient-norm clipping to the Muon param group (e.g. Megatron-LM `ChainedOptimizer`), the clip coefficient can scale per-matrix gradients below this floor → Newton–Schulz silently emits degenerate updates → training stalls while the forward/loss look completely normal, which is very hard to diagnose.

Fixes #229

### DIFF
--- a/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
+++ b/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
@@ -193,13 +193,8 @@ def newton_schulz(
         x = x.mT
 
     # Ensure spectral norm is at most 1.
-    # NOTE: ``eps`` is purely a divide-by-zero guard and must stay well below any realistic
-    # ``||x||_F`` (input is fp32). If it is too large, a small-norm input is divided by ``eps``
-    # instead of its norm, so ``||X||_F = ||x||_F / eps << 1`` and the iteration (tuned for
-    # singular values ~1) cannot lift it -> silently degenerate, non-orthogonal output. This
-    # breaks the scale-invariance of orthogonalization. It must also stay above ~1e-15 so that
-    # ``eps**2`` is still representable in fp32 (1e-15**2 = 1e-30, a normal float); a much smaller
-    # guard such as 1e-30 would underflow when squared. See issue #229.
+    # NOTE: ``eps`` is a divide-by-zero guard; it must stay well below any realistic ``||x||_F``
+    # yet remain fp32-safe when squared. See issue #229.
     if tp_group is not None:
         X = distributed_normalize_p2(x, eps, tp_group)
     else:
--- a/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
+++ b/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
@@ -141,7 +141,7 @@ def newton_schulz(
     steps: int,
     coefficient_type: NSCoeffT = "quintic",
     custom_coefficient_sets: list[tuple[float, float, float]] | None = None,
-    eps: float = 1e-7,
+    eps: float = 1e-30,
     transpose: bool | None = None,
     tp_group: torch.distributed.ProcessGroup | None = None,
     use_syrk: bool = False,
@@ -192,7 +192,12 @@ def newton_schulz(
     if transpose:
         x = x.mT
 
-    # Ensure spectral norm is at most 1
+    # Ensure spectral norm is at most 1.
+    # NOTE: ``eps`` is purely a divide-by-zero guard and must stay well below any realistic
+    # ``||x||_F`` (input is fp32). If it is too large, a small-norm input is divided by ``eps``
+    # instead of its norm, so ``||X||_F = ||x||_F / eps << 1`` and the iteration (tuned for
+    # singular values ~1) cannot lift it -> silently degenerate, non-orthogonal output. This
+    # breaks the scale-invariance of orthogonalization. See issue #229.
     if tp_group is not None:
         X = distributed_normalize_p2(x, eps, tp_group)
     else:
--- a/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
+++ b/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
@@ -141,7 +141,7 @@ def newton_schulz(
     steps: int,
     coefficient_type: NSCoeffT = "quintic",
     custom_coefficient_sets: list[tuple[float, float, float]] | None = None,
-    eps: float = 1e-30,
+    eps: float = 1e-15,
     transpose: bool | None = None,
     tp_group: torch.distributed.ProcessGroup | None = None,
     use_syrk: bool = False,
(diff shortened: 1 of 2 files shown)


yuchenwang3:
**Background / motivation:** surfaced while running **ms-swift + Megatron-Core SFT of Qwen3.5-35B-A3B** (GatedDeltaNet hybrid, `optimizer=dist_muon`) on **16× B200 across 2 nodes (2×8)**. A benign-but-large global `grad_norm` made the framework's gradient-clip coefficient tiny (`≈2e-8`), which pushed Muon's per-matrix gradients below this `F.normalize(eps=1e-7)` floor → `newton_schulz` silently returned a degenerate, non-orthogonal update → training stalled with forward/loss looking normal. The hard-to-diagnose part was exactly the *silent* degeneration, hence this scale-robustness fix. Megatron-side counterpart (the clip that produces the tiny coefficient): NVIDIA/Megatron-LM#5394 and PR NVIDIA/Megatron-LM#5395.

emerging_optimizers/orthogonalized_optimizers/muon_utils.py
```diff
@@ -192,7 +192,14 @@ def newton_schulz(
     if transpose:
         x = x.mT
 
-    # Ensure spectral norm is at most 1
+    # Ensure spectral norm is at most 1.
+    # NOTE: ``eps`` is purely a divide-by-zero guard and must stay well below any realistic
+    # ``||x||_F`` (input is fp32). If it is too large, a small-norm input is divided by ``eps``
+    # instead of its norm, so ``||X||_F = ||x||_F / eps << 1`` and the iteration (tuned for
+    # singular values ~1) cannot lift it -> silently degenerate, non-orthogonal output. This
+    # breaks the scale-invariance of orthogonalization. It must also stay above ~1e-15 so that
```
skyw: Statement is too strong. a reference to #229 should be sufficientl

tests/test_muon_utils.py
```diff
@@ -120,6 +120,29 @@ def test_newtonschulz5_close_to_reference(self, dim1, dim2):
             rtol=1e-7,
         )
 
+    @parameterized.parameters(1e-2, 1e-6, 1e-9, 1e-12)
+    def test_newtonschulz_scale_invariance(self, scale):
```
skyw: Test name is not quite right. I'd use more explicit name like test small eps etc.

skyw (changes requested):
eps change LGTM. rest is not necessary.

yuchenwang3:
Done in 620196c — both addressed:
- Trimmed the NOTE down to a one-line `See issue #229` reference.
- Renamed the test to `test_newtonschulz_small_eps`.

eps=1e-15 kept as you OK'd. Ready to merge whenever the re-run is green.

skyw (approved):

PR #235, opened 2026-06-24
Add stacked soap

skyw:
Save memory on grouped linear layers for MoE. Accuracy impact is yet to be tested.


FDecaYed (approved):
there is also a 2D check in _init_group which would fail if called on 3D weight. we need to fix that before exposing to megatron. otherwise looks good

skyw:
> there is also a 2D check in _init_group which would fail if called on 3D weight. we need to fix that before exposing to megatron. otherwise looks good

Yeah, shape manipulation on top of megatron distributed checkpoint needs more consideration. Need to test accuracy of this one first.

PR #239, opened 2026-07-01
Forward use_syrk through newton_schulz_tp

Wohox:
Passing use_syrk argument to newton_schulz_tp. Megatron only uses the tp version even if tp is not used (by passing None to tp_group). In order for megatron to use syrk kernels, the argument needs to be exposed.

Low risk change. Merge without test to save time.

### DIFF
--- a/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
+++ b/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
@@ -253,6 +253,7 @@ def newton_schulz_tp(
     tp_group: torch.distributed.ProcessGroup,
     partition_dim: int | None = None,
     tp_mode: Literal["duplicated", "distributed"] = "duplicated",
+    use_syrk: bool = False,
 ) -> torch.Tensor:
     """Tensor Parallel Newton-Schulz iteration.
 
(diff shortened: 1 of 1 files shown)


skyw:
Is the description AI written? it is too verbose and hard to catch the point. It it was, please rewrite in more concise form.

The benchmarked case also doesn't really show benefit. As it is probably more communication limited. The code itself is fine, but could use some real case to justify.

Test is missing. We also don't have infrastructure to test it other than smoke test. We can try to enable multiple GPU test environment. Given it is a simple change, I think we can accept with a smoke test on CPU just reach the triton kernel reports error.

And the plan was to use cuSolverMp version through TE integration, which is supposed to be faster.

Wohox:
> Is the description AI written? it is too verbose and hard to catch the point. It it was, please rewrite in more concise form.
> 
> The benchmarked case also doesn't really show benefit. As it is probably more communication limited. The code itself is fine, but could use some real case to justify.
> 
> Test is missing. We also don't have infrastructure to test it other than smoke test. We can try to enable multiple GPU test environment. Given it is a simple change, I think we can accept with a smoke test on CPU just reach the triton kernel reports error.
> 
> And the plan was to use cuSolverMp version through TE integration, which is supposed to be faster.

Yes this is AI generated description. Will make it more concise.
By the way, is there any analysis on how much does syrk and cuSolverMp should help with the throughput?

@gdengk Do we have a sound reason for enabling syrk here, looking at the benchmark results here, it seems this feature hasn't improved performance.

FDecaYed (approved):
LGTM

gdengk:
> > Is the description AI written? it is too verbose and hard to catch the point. It it was, please rewrite in more concise form.
> > The benchmarked case also doesn't really show benefit. As it is probably more communication limited. The code itself is fine, but could use some real case to justify.
> > Test is missing. We also don't have infrastructure to test it other than smoke test. We can try to enable multiple GPU test environment. Given it is a simple change, I think we can accept with a smoke test on CPU just reach the triton kernel reports error.
> > And the plan was to use cuSolverMp version through TE integration, which is supposed to be faster.
> 
> Yes this is AI generated description. Will make it more concise. By the way, is there any analysis on how much does syrk and cuSolverMp should help with the throughput?
> 
> @gdengk Do we have a sound reason for enabling syrk here, looking at the benchmark results here, it seems this feature hasn't improved performance.

For DSV4 pro case, Muon computation is taking 3% e2e for TP1EP64 config. So SYRK could potential help reduce the total computation time and also no extra comm cost with TP1 here. For some reason, it's calling TPmuon module instead of the other module already with SYRK. Not sure if we should route it to nonTP module for this TP=1 case?
I am currently running tests to see how much Si gain from SYRK triton kernel. Agreed that cublasSolver kernel will be the long term solution. @skyw do we have an idea when the cublasSolver would be integrated?

skyw:
@gdengk reached out and explained the motivation. Since it is a low risk change, let's merge as it is to save time.

@Wohox I update the description to reflect what the PR does. Try not to submit AI generated verbose message if not many (if at all) will read them. Also please add test in a follow up PR immediately.

PR #241, opened 2026-07-03
Add CPU smoke test for use_syrk in newton_schulz_tp

Wohox:
Follow-up to #239, per @skyw's request to add a test.

Adds `TestTensorParallelNewtonSchulz.test_use_syrk_reaches_triton_kernel` in `test_distributed_muon_utils_cpu.py`. Using a real gloo process group, it verifies `use_syrk=True` is wired through `newton_schulz_tp` to the Triton SYRK kernel (both `tp_mode`s, both partition dims): on CPU the kernel can't run, so reaching it raises (`assertRaisesRegex` on the kernel-path errors); `use_syrk=False` stays on the dense path and succeeds.

Verified on CPU via torchrun (gloo) at the ranks CI uses:
`torchrun --nproc_per_node=8 --no-python python tests/test_distributed_muon_utils_cpu.py TestTensorParallelNewtonSchulz --device=cpu -v -2` → passes (also at 4 and 2).

### DIFF
--- a/tests/test_distributed_muon_utils_cpu.py
+++ b/tests/test_distributed_muon_utils_cpu.py
@@ -21,6 +21,7 @@ from _comparison import assert_equal
 from absl import flags, logging
 from absl.testing import absltest, parameterized
 
+from emerging_optimizers import utils
 from emerging_optimizers.orthogonalized_optimizers import muon_utils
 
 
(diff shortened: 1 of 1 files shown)


tests/test_muon_utils.py
```diff
@@ -463,5 +463,33 @@ def test_match_newton_schulz_step_by_gemm(self, dim1, dim2):
         assert_equal(test_out, test_ref)
 
 
+@absltest.skipIf(
+    torch.cuda.is_available(),
+    "CPU-only smoke test: on a GPU the Triton SYRK kernel actually runs and would not error.",
+)
+class TestNewtonSchulzTpUseSyrkCpuSmoke(parameterized.TestCase):
```
skyw: I meant to have it in test_distributed_muon_utils_cpu.py and test with real process group (gloo on CPU).

tests/test_muon_utils.py
```diff
@@ -463,5 +463,33 @@ def test_match_newton_schulz_step_by_gemm(self, dim1, dim2):
...
+    def test_use_syrk_reaches_triton_kernel_and_errors_on_cpu(self):
+        x = torch.randn(16, 32, dtype=torch.float32)
+        with utils.fp32_matmul_precision("medium"):
+            # use_syrk=False stays on the dense Newton-Schulz path and succeeds on CPU.
+            muon_utils.newton_schulz_tp(
+                x, steps=5, coefficient_type="quintic", tp_group=None, partition_dim=None, use_syrk=False
+            )
+            # use_syrk=True routes newton_schulz_tp -> newton_schulz -> the Triton SYRK kernel, which
+            # cannot run on CPU (no GPU driver, or an older Triton without the tensor-descriptor API).
+            with self.assertRaises((RuntimeError, AssertionError)):
```
skyw: Should also capture specific error message so that we know it has reached expected places.

skyw (changes requested):

Wohox:
Thanks — addressed both:
- Moved the test into `test_distributed_muon_utils_cpu.py` and it now runs against a real gloo process group, covering both `tp_mode`s (`distributed`, `duplicated`) and both partition dims.
- Switched to `assertRaisesRegex((RuntimeError, AssertionError), r"active drivers|tensor descriptor")` so the failure must come from the SYRK kernel path (triton driver check / tensor-descriptor assert), not an unrelated earlier error.

Verified via torchrun (2 ranks, gloo, CPU) in the NGC-equivalent container: `TestTensorParallelNewtonSchulz` → 14 passed.

tests/test_distributed_muon_utils_cpu.py
```diff
@@ -227,6 +228,59 @@ def test_1step_close_to_non_distributed(self, shape, partition_dim, tp_mode):
 
         torch.testing.assert_close(ref_out.chunk(world_size, dim=partition_dim)[rank], dist_out, atol=1e-6, rtol=0)
 
+    @parameterized.product(
+        partition_dim=(0, 1),
+        tp_mode=("distributed", "duplicated"),
+    )
+    def test_use_syrk_reaches_triton_kernel(self, partition_dim, tp_mode):
+        """Smoke test that use_syrk is wired through newton_schulz_tp to the Triton SYRK kernel.
```
skyw: Nit: remove meaningless docstring generated by AI. the test name is self-explainary enough.

skyw (dismissed):

Wohox:
Removed the docstring and the inline comments — the test name is self-explanatory (thanks @skyw ~)

tests/test_muon_utils.py
Wohox: added specific error message~

skyw (approved):

PR #245, opened 2026-07-09
fix(syrk): fall back to GEMM Newton-Schulz for non-16B-aligned matrices

gdengk:
SYRK's `tsyrk_ex` uses TMA descriptors requiring 16-byte-aligned strides, so both matrix dims must be multiples of 8; unaligned weights (e.g. DSv4 MHC `(4, N)`) assert "strides must be 16-byte aligned".
Guard the `newton_schulz` SYRK dispatch to take the TMA path only when both dims are multiples of 8, otherwise fall back to the plain GEMM Newton-Schulz step.

### DIFF
--- a/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
+++ b/emerging_optimizers/orthogonalized_optimizers/muon_utils.py
@@ -249,7 +249,19 @@ def newton_schulz(
         if use_syrk:
             if X.ndim > 2:
                 raise TypeError("use_syrk does not support N-d input.")
-            ns_step_fn = newton_schulz_step_tsyrk
+            # The SYRK Triton kernel uses TMA TensorDescriptors, which require 16-byte-aligned row
+            # strides, i.e. both matrix dims must be multiples of 8 for bf16. Unaligned shapes (e.g.
+            # DeepSeek-V4 Hyper-Connections weights with dim=num_residual_streams=4, whose (4, 4)
+            # product has an 8-byte row stride) would raise "strides must be 16-byte aligned"; keep
+            # the plain GEMM Newton-Schulz step instead.
+            if X.size(-1) % 8 == 0 and X.size(-2) % 8 == 0:
+                ns_step_fn = newton_schulz_step_tsyrk
+            else:
+                logging.log_first_n(
+                    logging.WARNING,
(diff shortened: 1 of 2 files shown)


emerging_optimizers/orthogonalized_optimizers/muon_utils.py
```diff
@@ -249,7 +249,20 @@ def newton_schulz(
         if use_syrk:
             if X.ndim > 2:
                 raise TypeError("use_syrk does not support N-d input.")
-            ns_step_fn = newton_schulz_step_tsyrk
+            # The SYRK Triton kernel uses TMA TensorDescriptors, which require 16-byte-aligned row
```
skyw: Comment is too verbose and start to defeat the goal. More people will skip long comment instead of reading it and understand what it is trying to say.
"# The SYRK Triton kernel uses TMA TensorDescriptors, which require 16-byte-aligned row # strides, i.e. both matrix dims must be multiples of 8 for bf16. Fallback to gemm if not satisfied."
is more than enough.

And it is always better to let code speak itself. For example in this case, a more clear logging message would also suffice.

emerging_optimizers/orthogonalized_optimizers/muon_utils.py
```diff
@@ -249,7 +249,20 @@ def newton_schulz(
...
+            # strides, i.e. both matrix dims must be multiples of 8 for bf16. Unaligned shapes (e.g.
+            # DeepSeek-V4 Hyper-Connections weights with dim=num_residual_streams=4, whose (4, 4)
+            # product has an 8-byte row stride) would raise "strides must be 16-byte aligned"; keep
+            # the plain GEMM Newton-Schulz step instead.
+            if X.size(-1) % 8 == 0 and X.size(-2) % 8 == 0:
+                ns_step_fn = newton_schulz_step_tsyrk
+            else:
+                logging.log_first_n(
+                    logging.WARNING,
+                    "[SYRK-SKIP] non-8-aligned shape %s; using GEMM Newton-Schulz.",
```
skyw: Let's use logging.ERROR level for fallbacks.

Also ambitious message. can say either 16-bytes aligned, or simply m and n must be multiple of 8 in bf16

skyw (changes requested):
Would also be good to add a test to make sure the guard worked.

skyw (approved):
The test logic is not super straightforward, but serves the purpose.

Issue #188, opened 2026-05-20
Adding symmetry-compatible / equivariant optimizers for general matrix-valued parameters


timlautk:
**Is your feature request related to a problem? Please describe.**
I would like to request support for symmetry-compatible / equivariant optimizers for general matrix-valued parameters, proposed in [Symmetry-Compatible Principle for Optimizer Design: Embeddings, LM Heads, SwiGLU MLPs, and MoE Routers](https://arxiv.org/abs/2605.18106).

Many modern neural-network parameters are naturally matrices rather than unstructured vectors: attention projections, MLP projections, embeddings, LM heads, and MoE routers. Standard coordinate-wise optimizers such as AdamW treat these matrices as collections of independent scalar coordinates, which can break natural symmetries such as left/right orthogonal equivariance, vocabulary permutation equivariance, or expert permutation equivariance.

Reference implementation:
https://github.com/timlautk/equivariant_optimizers

**Describe the solution you'd like**
I would like the optimizer API to support assigning equivariant optimizer rules to matrix-valued parameters based on their layer type and symmetry structure.

Concretely, this could include support for optimizer variants such as:

- **Full spectral / polar-gradient updates** for ordinary dense matrix parameters, such as attention and linear layers.
- **Right-spectral updates** for embedding and LM-head matrices, respecting vocabulary permutation symmetry and hidden-feature orthogonal symmetry.
- **Row-norm or hybrid row-norm/right-spectral updates** for large vocabulary-indexed matrices.
- **Left-spectral or centered equivariant updates** for MoE router matrices, respecting expert permutation symmetry and logit-shift structure.
- Momentum variants such as EMA or momentum-first polar updates.

Ideally, the implementation would allow users to define parameter groups by module type or tensor shape and attach different symmetry-compatible update rules to different matrix classes.

**Describe alternatives you've considered**
An alternative is to use existing matrix optimizers such as Muon or Shampoo on some matrix parameters while leaving embeddings, LM heads, and routers on AdamW. This is useful in practice, but it does not provide a unified layerwise principle for assigning optimizers according to parameter symmetry.

A third alternative is to maintain this as an external optimizer package. However, native or officially supported integration would make it much easier to use these methods in large-scale training workflows.

**Additional context**
[...]

PR #247, opened 2026-07-10
Add left polargrad function

abtonmoy:
## What does this PR do?

Adds `left_polargrad_orth_fn`, the left-spectral counterpart of `right_polargrad_orth_fn` (#234), as a follow-up to #188:

$$u = (G G^\top)^{-1/2} G, \qquad \text{update} = \lVert G \rVert_*^{\alpha} \, u$$

Intended for wide matrices such as MoE router weights (`num_experts x hidden`), where `center_rows=True` respects expert-permutation symmetry and the logit-shift invariance of the routing softmax. Same shape as #234: a composable orthogonalization function plus a docstring example of building an optimizer from it, no new optimizer class and no new arguments.

## Why the inverse sqrt uses a pseudo-inverse cutoff

With `center_rows=True` the left Gram $G G^\top$ is always exactly singular: centering annihilates the all-ones direction. Mirroring the right variant's `clamp_min(eps).rsqrt()` there turns eigh noise on the null eigenvalue into ~`eps^{-1/2}` amplification; across 600 randomized fp32 trials the worst spurious error in the permutation-equivariance check was ~3.5e2, on the order of the update itself. The right variant does not hit this because its Gram stays full-rank for tall inputs.

This implementation drops eigendirections below `eigvals.amax() * nrows * fp32_eps` instead. On the same 600 trials the worst error becomes 3.2e-3 (2e-4 relative), results on full-rank inputs are bit-identical to the clamp formula, and the same treatment covers uncentered inputs with more rows than columns.

## Tests

Mirrors the #234 test style:

- Left-orthogonal equivariance `f(QG) = Q f(G)`.
- Row-permutation equivariance with and without `center_rows`. Centering only commutes with transformations that fix the all-ones vector, so the centered case is tested with permutations rather than general orthogonal matrices, and uses looser tolerances (fp32 eigh noise on the singular Gram dominates the comparison).
- Centered updates keep zero column means.
- Composition through `OrthogonalizedOptimizer.step()`.

Verified locally: `test_polargrad.py` and `test_orthogonalized_optimizer.py` pass on `--device=cpu` and `--device=cuda`, random seed and `--seed=42`; mypy and pre-commit clean.

### DIFF
--- a/docs/apidocs/orthogonalized-optimizers.md
+++ b/docs/apidocs/orthogonalized-optimizers.md
@@ -46,6 +46,8 @@ emerging_optimizers.orthogonalized_optimizers
 .. autoclass:: PolarGrad
     :members:
 
+.. autofunction:: left_polargrad_orth_fn
+
 .. autofunction:: right_polargrad_orth_fn
 
 
--- a/emerging_optimizers/orthogonalized_optimizers/polargrad.py
+++ b/emerging_optimizers/orthogonalized_optimizers/polargrad.py
@@ -26,7 +26,7 @@ from emerging_optimizers.utils import FP32MatmulPrecT
 from emerging_optimizers.utils.eig import eigh_with_fallback
 
 
-__all__ = ["PolarGrad", "right_polargrad_orth_fn"]
+__all__ = ["PolarGrad", "left_polargrad_orth_fn", "right_polargrad_orth_fn"]
 
 
 @registry.register_optimizer("polargrad")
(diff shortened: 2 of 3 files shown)


skyw (approved):
Thanks for the contribution, let's merge it for now.

Could use a consolidation to make left/right in the same function.

abtonmoy:
Thanks for the review! Happy to do the consolidation as a follow-up. I'd fold both into a single one-sided polar function, e.g. `one_sided_polargrad_orth_fn(grad, side="left"|"right")`, sharing the Gram/eigh/pseudo-inverse path (the cutoff also protects the right variant for square-ish inputs), and keep `left_polargrad_orth_fn`/`right_polargrad_orth_fn` as thin partials so existing callers and the docs stay stable. Let me know if you'd rather drop the wrappers entirely.

After that I was planning to continue the #188 follow-ups with a row-norm orthogonalization function and a hybrid (polar plus row-scaling) variant in the same composable style, which I'll open separately.

timlautk:
> Thanks for the review! Happy to do the consolidation as a follow-up. I'd fold both into a single one-sided polar function, e.g. `one_sided_polargrad_orth_fn(grad, side="left"|"right")`, sharing the Gram/eigh/pseudo-inverse path (the cutoff also protects the right variant for square-ish inputs), and keep `left_polargrad_orth_fn`/`right_polargrad_orth_fn` as thin partials so existing callers and the docs stay stable. Let me know if you'd rather drop the wrappers entirely.
> 
> After that I was planning to continue the #188 follow-ups with a row-norm orthogonalization function and a hybrid (polar plus row-scaling) variant in the same composable style, which I'll open separately.

Really cool! Thanks so much for the contributions!

PR #248, opened 2026-07-11
Consolidate left/right polargrad into one function

abtonmoy:
## What does this PR do?

Follow-up to the #247 review suggestion to consolidate left/right into the same function. Adds `one_sided_polargrad_orth_fn(grad, side="left"|"right")` holding the shared Gram/eigh/pseudo-inverse implementation, and reduces `left_polargrad_orth_fn` / `right_polargrad_orth_fn` to thin wrappers so existing callers and docs entries are unchanged. Invalid `side` raises `ValueError`. Happy to drop the wrappers instead if preferred.

## Behavior note

The shared path uses the pseudo-inverse cutoff introduced in #247 for the left side. On tall full-rank inputs the right side is bit-identical to the previous formula (verified over 800 randomized cases). On rank-deficient inputs (wide matrices, or centered square ones) the previous right formula amplified eigh noise through `clamp_min(1e-15).rsqrt()`, producing updates up to ~3e4 in magnitude; the shared path drops those null directions and stays at the nuclear-norm scale.

## Tests

- `test_invalid_side_raises_value_error`
- `test_wrapper_output_matches_one_sided` (bitwise, both sides)
- `test_left_close_to_transposed_right` (mirror identity `f_left(G) == f_right(G^T)^T`)
- Existing left/right test classes are unchanged and now exercise the wrappers end-to-end.

Verified locally: `test_polargrad.py` (34 tests) and `test_orthogonalized_optimizer.py` (143 tests) pass on `--device=cpu` and `--device=cuda`, random seed and `--seed=42`; pre-commit (ruff, mypy, format) clean.

### DIFF
--- a/emerging_optimizers/orthogonalized_optimizers/polargrad.py
+++ b/emerging_optimizers/orthogonalized_optimizers/polargrad.py
@@ -129,49 +129,12 @@ def one_sided_polargrad_orth_fn(
     ``side="right"`` suits tall matrices (e.g. an embedding or LM-head weight, ``vocab x hidden``);
     ``side="left"`` suits wide matrices (e.g. an MoE router weight, ``num_experts x hidden``).
 
-    .. code-block:: python
-       :caption: Define a ``LeftPolarGrad`` by partially applying this orthogonalization
-
-       class LeftPolarGrad(OrthogonalizedOptimizer):
-           def __init__(
-               self,
-               params,
-               lr: float = 3e-4,
-               momentum: float = 0.95,
-               weight_decay: float = 0.01,
-               *,
-               ...
-               alpha: float = 1.0,
-               center_rows: bool = False,
-               eps: float = 1e-15,
-               extra_scale_factor: float = 1.0,
-           ) -> None:
-               scaled_orthogonalize_fn = functools.partial(
-                   one_sided_polargrad_orth_fn,
-                   side="left",
-                   alpha=alpha,
-                   center_rows=center_rows,
-                   eps=eps,
-                   extra_scale_factor=extra_scale_factor,
-               )
-               super().__init__(
-                   ...
-               )
-
-    Note:
-        The inverse square root is computed as a pseudo-inverse: eigendirections of the Gram matrix that
-        are numerically indistinguishable from its null space contribute nothing to the update instead
-        of being amplified by the ``eps`` floor. This matters because the Gram is singular whenever
-        ``center_rows=True`` on the ``side="left"`` Gram (centering annihilates the all-ones direction)
-        or the matrix is rank deficient.
-
     Args:
         grad: The (momentum) tensor to orthogonalize.
         side: Which polar factor to orthogonalize, ``"left"`` or ``"right"``.
         alpha: Exponent applied to the nuclear-norm scale factor.
         center_rows: If True, subtract the per-column mean (the average over the row axis, ``dim=0``)
-            before and after the update, so each column is zero-mean. For MoE routers this respects the
-            logit-shift invariance of the routing softmax.
+            before and after the update, so each column is zero-mean.
         eps: Floor on the Gram eigenvalues for the nuclear-norm computation.
         extra_scale_factor: Extra multiplier on the update.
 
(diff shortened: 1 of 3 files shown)


tests/test_polargrad.py
```diff
@@ -196,5 +196,42 @@ def test_usable_as_scaled_orthogonalize_fn(self) -> None:
...
+        """left/right wrappers produce bitwise-identical results to side= selection."""
+        grad = torch.randn((8, 32) if side == "left" else (32, 8), device=self.device)
+        wrapper = polargrad.left_polargrad_orth_fn if side == "left" else polargrad.right_polargrad_orth_fn
+
+        assert_equal(
+            wrapper(grad, alpha=1.0, center_rows=True, extra_scale_factor=0.2),
+            polargrad.one_sided_polargrad_orth_fn(
+                grad, side=side, alpha=1.0, center_rows=True, extra_scale_factor=0.2
+            ),
+        )
```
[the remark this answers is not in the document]
skyw: @greptileai append the rule, synchronizations in test assertion is ok.

emerging_optimizers/orthogonalized_optimizers/polargrad.py
```diff
@@ -150,26 +158,44 @@ def __init__(
                    ...
                )
 
+    Note:
```
skyw: Too verbose, doesn't really say anything more than the equation and the comment of eps. can just delete it.

emerging_optimizers/orthogonalized_optimizers/polargrad.py
```diff
@@ -150,26 +158,44 @@ def __init__(
...
+        or the matrix is rank deficient.
+
     Args:
         grad: The (momentum) tensor to orthogonalize.
+        side: Which polar factor to orthogonalize, ``"left"`` or ``"right"``.
         alpha: Exponent applied to the nuclear-norm scale factor.
-        center_rows: If True, subtract the per-column mean (the average over the row / vocabulary axis,
-            ``dim=0``) before and after the update, so each column is zero-mean.
-        eps: Floor on the right-Gram eigenvalues for the inverse sqrt and nuclear-norm computation.
+        center_rows: If True, subtract the per-column mean (the average over the row axis, ``dim=0``)
```
skyw, line 136: This function doesn't make any assumption that how MoE routers or any other module is implemented. rollback wrong comment change.

emerging_optimizers/orthogonalized_optimizers/polargrad.py
```diff
@@ -150,26 +158,44 @@ def __init__(
...
     m = grad.to(torch.float32)
     if center_rows:
         m = m - m.mean(dim=0, keepdim=True)
 
-    eigvals, eigvecs = eigh_with_fallback(m.transpose(-1, -2) @ m)
+    gram = m @ m.transpose(-1, -2) if side == "left" else m.transpose(-1, -2) @ m
+    eigvals, eigvecs = eigh_with_fallback(gram)
     eigvals.clamp_min_(eps)
-    right_gram_inv_sqrt = (eigvecs * eigvals.rsqrt().unsqueeze(-2)) @ eigvecs.transpose(-1, -2)
+    cutoff = eigvals.amax() * gram.shape[-1] * torch.finfo(torch.float32).eps
```
skyw: Do not use floating format eps.

emerging_optimizers/orthogonalized_optimizers/polargrad.py
```diff
@@ -178,83 +204,35 @@ def __init__(
     return update.to(grad.dtype)
 
 
-def left_polargrad_orth_fn(
+def right_polargrad_orth_fn(
```
skyw, line 170: Inconsistent with examples in `one_sided_polargrad_orth_fn`'s docstring. Can either
* delete the example, it is self-explainary enough.
* or use partial to define left_ and right_ function.

skyw (changes requested):
Largely ok, some minor things need to be fixed.

abtonmoy:
Addressed, thanks

skyw (approved):

PR #253, opened 2026-07-15
Add row norm function

abtonmoy:
## What does this PR do?

Adds `row_norm_fn`, a composable update normalization from #188: each row of the momentum is scaled to unit norm, with optional `center_rows` matching the polargrad functions. Usable as `scaled_orthogonalize_fn` of `OrthogonalizedOptimizer`; no new optimizer class.

Placed in `orthogonalized_optimizers` for composability. As noted in #188 this is not strictly an orthogonalization, so happy to move it if you prefer another home.

## Tests

Unit row norms, per-row scale invariance, row-permutation equivariance, right-orthogonal equivariance, centered-output projection, zero-gradient edge case, and composition through `OrthogonalizedOptimizer.step()`. Verified locally on `--device=cpu` and `--device=cuda`, random seed and `--seed=42`; pre-commit clean.


emerging_optimizers/contrib/row_norm.py
```diff
@@ -0,0 +1,57 @@
...
+    Args:
+        grad: The (momentum) tensor to normalize.
+        center_rows: If True, subtract the per-column mean (the average over the row axis, ``dim=0``)
+            before and after the update, so each column is zero-mean.
+        eps: Floor on the row norms.
+        extra_scale_factor: Extra multiplier on the update.
+
+    Returns:
+        The row-normalized update, same shape and dtype as ``grad``.
+    """
```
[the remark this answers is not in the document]
skyw: Seems a valid concern worth fixing.

emerging_optimizers/contrib/row_norm.py
```diff
@@ -0,0 +1,57 @@
...
+    m = grad.to(torch.float32)
+    if center_rows:
+        m = m - m.mean(dim=0, keepdim=True)
```
skyw, line 58: Guard shape of input grad, at least document.

tests/test_row_norm.py
```diff
@@ -0,0 +1,123 @@
...
+        torch.testing.assert_close(
+            update.norm(dim=-1),
+            torch.ones(num_rows, device=self.device),
+        )
+
+    @parameterized.parameters((8, 4), (8, 32))
+    def test_row_scale_invariance(self, num_rows, num_cols) -> None:
+        """f(D G) == f(G) for a positive per-row scaling D."""
+        grad = torch.randn((num_rows, num_cols), device=self.device)
+        scales = torch.rand((num_rows, 1), device=self.device) + 0.5
```
skyw: what's the reason to add +0.5? -0.5 can make it zero centered.
It is better to use integer or power of two number for the scales to minimize difference caused by floating point arithetimic.

skyw (commented):
Code looks ok. 
Could you move it under contrib. We are making some adjustment to project layout.

abtonmoy:
Addressed, thanks

skyw (approved):

PR #258, opened 2026-07-24
fix: allow Muon to run without Triton

returnL:
## Summary

- Make Triton optional when importing Muon.
- Fall back to the regular Newton-Schulz implementation when Triton is unavailable.
- Add a regression test that simulates an environment without Triton.

## Motivation

Triton is not a runtime dependency, and Muon defaults to `use_syrk=False`. However, `triton_kernels` currently imports the SYRK implementation unconditionally, causing Muon imports to fail with `ModuleNotFoundError` when Triton is not installed.

This change catches only `ModuleNotFoundError` raised for the top-level `triton` package and sets `HAS_TRITON_340` to `False`. Other import errors are still propagated.

There is no behavior change when Triton is installed. When Triton is unavailable, requesting `use_syrk=True` follows the existing fallback behavior and uses the regular Newton-Schulz implementation.

## Testing

- `python tests/test_optional_triton.py --device=cpu`
- `ruff check emerging_optimizers/triton_kernels/__init__.py tests/test_optional_triton.py`
- `ruff format --check emerging_optimizers/triton_kernels/__init__.py tests/test_optional_triton.py`
- `git diff --check`


skyw:
Looks extend the guard in syrk.py should be enough? which minimizes code change.

skyw:
> Looks extend the guard in syrk.py should be enough? which minimizes code change.

Actually, syrk assumes triton exist. Need to guard it higher up.

skyw (changes requested):
By default we test in NGC pytorch container which always has triton.

Import guard part LGTM. The way test is constructed can't be accepted by out CI.

We can actauly accept this change without the tests.

returnL:
Thanks for the review. I removed the subprocess-based test as suggested and kept the import guard at the package level. The PR now only contains the import guard change.

skyw (approved):

PR #260, opened 2026-07-24
fix: increment step counter in PSGDPro

andrewwhitecdw:
## Bug

`state["step"]` was initialized to 0 and passed to `_get_precond_lr`, but it was never incremented. Every optimization step therefore used `step=0`, so the preconditioner learning-rate schedule never progressed.

## Fix

Increment `state["step"]` at the end of each per-parameter update loop, consistent with SOAP/MOSO/TpRekls.

## Test

Added `tests/test_psgd_pro.py::TestPSGDPro::test_step_counter_increments`, which verifies the step counter advances across multiple optimizer steps.

## Verification

`python tests/test_psgd_pro.py` passes.
`ruff check emerging_optimizers/psgd/psgd.py tests/test_psgd_pro.py` and `ruff format --check ...` pass.

### DIFF
--- a/tests/test_psgd_pro.py
+++ b/tests/test_psgd_pro.py
@@ -1,0 +1,52 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


tests/test_psgd_pro.py
```diff
@@ -0,0 +1,52 @@
...
+class TestPSGDPro(absltest.TestCase):
+    def setUp(self) -> None:
+        self.device = FLAGS.device
+
+    def test_step_counter_increments(self) -> None:
+        """PSGDPro must increment state['step'] each optimization step."""
+        param = torch.randn(4, 4, device=self.device, requires_grad=True)
+        optimizer = PSGDPro([param], lr=0.01)
+
+        for expected_step in range(1, 4):
```
skyw, line 45: Nit: comment why starting loop count from 1 is expected_step.

skyw (approved):
LGTM.
Thanks for the contribution.

PR #261, opened 2026-07-24
fix: mutate exp_avg_fast in place when beta_fast is zero in AdEMAMix

andrewwhitecdw:
## Bug

When `beta_fast == 0.0`, the `else` branch reassigned the local variable `exp_avg_fast = grad` instead of updating the caller`s tensor in place. The optimizer state buffer was therefore left unchanged across steps.

## Fix

Use `exp_avg_fast.lerp_(grad, 1 - beta_fast)` unconditionally; it already behaves correctly when `beta_fast == 0.0` (`lerp_(grad, 1.0)` sets the buffer to `grad`).

## Test

Added `UpdateFunctionTest::test_calculate_ademamix_update_mutates_exp_avg_fast_when_beta_fast_zero`, which verifies the caller`s `exp_avg_fast` buffer is mutated.

## Verification

`python tests/test_scalar_optimizers.py UpdateFunctionTest.test_calculate_ademamix_update_mutates_exp_avg_fast_when_beta_fast_zero` passes.
`ruff check emerging_optimizers/scalar_optimizers/update_functions/ademamix.py tests/test_scalar_optimizers.py` and `ruff format --check ...` pass.

### DIFF
--- a/emerging_optimizers/scalar_optimizers/update_functions/ademamix.py
+++ b/emerging_optimizers/scalar_optimizers/update_functions/ademamix.py
@@ -93,10 +93,7 @@ def calculate_ademamix_update(
         bias_correction2 = 1
 
     # Decay the fast first moment, slow first moment and second moment with an exponential moving average
-    if beta_fast != 0.0:
-        exp_avg_fast.lerp_(grad, 1 - beta_fast)
-    else:
-        exp_avg_fast = grad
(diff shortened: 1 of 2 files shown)


skyw:
@mkhona-nvidia to review whether it was intentional or bug.

mkhona-nvidia (approved):

PR #262, opened 2026-07-27
added a light generalization of split qkv functionality to support pe…

mkhona-nvidia:
This is a small change of orthogonalize qkv function to support per-head orthogonalization


skyw:
This is a big change in terms of commitment. The optimizers we provided having no knowledge of type of the parameters. "qkv", "head" those are very implementation specific knowledge that optimizer can't safely make assumption upon so that they should not exist in arguments.

Supporting per head muon is not yet enough motivation to make a fundamental change.

PR #264, opened 2026-07-28
feat: add scaled CANS inverse root utility

mkhona-nvidia:
CANS polynomial-based root inverse computation from https://arxiv.org/abs/2506.10935

### DIFF
--- a/emerging_optimizers/soap/matrix_root_inverse_utils.py
+++ b/emerging_optimizers/soap/matrix_root_inverse_utils.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/emerging_optimizers/soap/matrix_root_inverse_utils.py
+++ b/emerging_optimizers/soap/matrix_root_inverse_utils.py
@@ -19,7 +19,7 @@ from emerging_optimizers import utils
 from emerging_optimizers.utils import FP32MatmulPrecT
 
 
-__all__ = ["scaled_cans_coupled_ns"]
+__all__ = ["mat_root_inv_via_scaled_cans"]
 
 # All constant arithmetic, including the 1.01 safety factor, is folded into these coefficients.
 _CANS_COEFFS = (
@@ -36,7 +36,7 @@ _CANS_COEFFS = (
 )
 
 
-def scaled_cans_coupled_ns(
+def mat_root_inv_via_scaled_cans(
     x: Tensor,
     eps: float = 1e-12,
     fp32_matmul_prec: FP32MatmulPrecT = "highest",
--- a/emerging_optimizers/soap/matrix_root_inverse_utils.py
+++ b/emerging_optimizers/soap/matrix_root_inverse_utils.py
@@ -89,12 +89,14 @@ def scaled_cans_coupled_ns(
         if torch.get_float32_matmul_precision() == "medium":
             y = y.to(torch.bfloat16)
 
-        z = torch.eye(x.shape[-1], device=x.device, dtype=y.dtype).expand_as(y)
-        cans_addmm = torch.addmm if x.dim() == 2 else torch.baddbmm
+        z = torch.eye(x.shape[-1], device=x.device, dtype=y.dtype)
+        if x.dim() == 3:
+            z = z.unsqueeze(0)
+
         for beta, alpha in _CANS_COEFFS:
             p = z @ y
-            z = cans_addmm(z, p, z, beta=beta, alpha=alpha)
-            y = cans_addmm(y, y, p, beta=beta, alpha=alpha)
+            z = torch.add(z * beta, p @ z, alpha=alpha)
+            y = torch.add(y * beta, y @ p, alpha=alpha)
 
         z = z.to(torch.float32)
         z.mul_(torch.rsqrt(inf_norm))
--- a/emerging_optimizers/soap/matrix_root_inverse_utils.py
+++ b/emerging_optimizers/soap/matrix_root_inverse_utils.py
@@ -1,0 +1,101 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import torch
+from torch import Tensor
+
+from emerging_optimizers import utils
+from emerging_optimizers.utils import FP32MatmulPrecT
+
+
+__all__ = ["scaled_cans_coupled_ns"]
+
+# All constant arithmetic, including the 1.01 safety factor, is folded into these coefficients.
+_CANS_COEFFS = (
+    (5.182503604966906, -5.126830178299687),
+    (2.586120737395915, -0.641538812403133),
+    (2.567364126726186, -0.6391058222170474),
+    (2.520560084348265, -0.6330225823828756),
+    (2.410759275435182, -0.6186815444268036),
+    (2.1883348130094173, -0.5893091162177136),
+    (1.8595760874873613, -0.5449991062102938),
+    (1.589020160467417, -0.5075811685214573),
+    (1.5051653981684994, -0.4957799077972079),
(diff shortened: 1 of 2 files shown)


emerging_optimizers/soap/matrix_root_inverse_utils.py
```diff
@@ -0,0 +1,74 @@
...
+    Returns:
+        The approximate inverse square root with the same shape as ``x``.
+    """
+    if x.dim() not in (2, 3) or x.shape[-2] != x.shape[-1]:
+        raise TypeError(f"x must be a square matrix or batch of square matrices, got shape {tuple(x.shape)}")
+    if not x.is_cuda:
+        raise TypeError("scaled_cans_coupled_ns only supports CUDA tensors")
+
+    with utils.fp32_matmul_precision(fp32_matmul_prec):
+        input_dtype = x.dtype
```
skyw: don't need this, use of fp32_matmul_precision context manager assumes input is in fp32. can cast to bf16 if precision is medium, same as what we do in muon.

emerging_optimizers/soap/matrix_root_inverse_utils.py
```diff
@@ -0,0 +1,74 @@
...
+        eps: Lower bound used when normalizing the matrices.
+        fp32_matmul_prec: Precision used for FP32 matrix multiplications: ``"medium"`` for BF16,
+            ``"high"`` for TF32, or ``"highest"`` for FP32.
+
+    Returns:
+        The approximate inverse square root with the same shape as ``x``.
+    """
+    if x.dim() not in (2, 3) or x.shape[-2] != x.shape[-1]:
+        raise TypeError(f"x must be a square matrix or batch of square matrices, got shape {tuple(x.shape)}")
+    if not x.is_cuda:
```
skyw: I think the code works on CPU. If not, could you check which part? I'd like to keep the tradition.

emerging_optimizers/soap/matrix_root_inverse_utils.py
```diff
@@ -0,0 +1,74 @@
...
+import torch
+from torch import Tensor
+
+from emerging_optimizers import utils
+from emerging_optimizers.utils import FP32MatmulPrecT
+
+
+__all__ = ["scaled_cans_coupled_ns"]
+
+_CANS_COEFFS = (
```
skyw, line 25: Add some comment that we folded all the constant arithmetic.

emerging_optimizers/soap/matrix_root_inverse_utils.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    if x.dim() not in (2, 3) or x.shape[-2] != x.shape[-1]:
+        raise TypeError(f"x must be a square matrix or batch of square matrices, got shape {tuple(x.shape)}")
+    if not x.is_cuda:
+        raise TypeError("scaled_cans_coupled_ns only supports CUDA tensors")
+
+    with utils.fp32_matmul_precision(fp32_matmul_prec):
+        input_dtype = x.dtype
+        x = x.float()
+        scale = torch.linalg.matrix_norm(x, ord=float("inf"), dim=(-2, -1), keepdim=True).clamp_min_(eps)
```
skyw: explicitly calling it `inf_norm` would be a better name here.

emerging_optimizers/soap/matrix_root_inverse_utils.py
```diff
@@ -0,0 +1,74 @@
...
+        y = x / scale
+        z = torch.eye(x.shape[-1], device=x.device, dtype=x.dtype).expand_as(x)
+        cans_addmm = torch.addmm if x.dim() == 2 else torch.baddbmm
+        for beta, alpha in _CANS_COEFFS:
+            p = z @ y
+            z, y = (
```
skyw: confusing syntax. just right them in separate lines

tests/test_matrix_root_inverse_utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+        self.assertEqual(inverse_root.shape, matrix.shape)
+        self.assertTrue(torch.isfinite(inverse_root).all())
+
+    def test_scaled_cans_rejects_cpu_tensor(self) -> None:
+        with self.assertRaisesRegex(TypeError, "only supports CUDA"):
+            scaled_cans_coupled_ns(torch.eye(4))
+
+    @parameterized.parameters("medium", "high", "highest")  # type: ignore[misc]
+    def test_scaled_cans_supports_fp32_matmul_precision(self, fp32_matmul_prec: FP32MatmulPrecT) -> None:
```
skyw: All of these just confirms it doesn't blow up. can combine them to a smoke test, and check output shape.

tests/test_matrix_root_inverse_utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+def setUpModule() -> None:
+    if FLAGS.seed is not None:
+        logging.info("Setting random seed to %d", FLAGS.seed)
+        torch.manual_seed(FLAGS.seed)
+        if torch.cuda.is_available():
+            torch.cuda.manual_seed_all(FLAGS.seed)
+
+
+class CANSUtilsTest(parameterized.TestCase):
```
skyw: Add a test against reconstructed input. with small sizes, they should be close enough.

skyw (changes requested):
logic looks ok. some minor things can be improved.

The only thin missing is a more robust test.

tests/test_matrix_root_inverse_utils.py
```diff
@@ -0,0 +1,70 @@
...
+    def test_scaled_cans_supports_optional_batch_dimension(self, shape: tuple[int, ...]) -> None:
+        if FLAGS.device != "cuda":
+            self.skipTest("CANS requires CUDA")
+
+        x = torch.randn(*shape, device=FLAGS.device)
+        matrix = x @ x.mT + 0.1 * torch.eye(shape[-1], device=FLAGS.device)
+
+        inverse_root = scaled_cans_coupled_ns(matrix)
+
+        self.assertEqual(inverse_root.shape, matrix.shape)
```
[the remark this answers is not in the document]
mkhona-nvidia: resolved in next test

emerging_optimizers/soap/matrix_root_inverse_utils.py
```diff
@@ -0,0 +1,101 @@
...
+        z = torch.eye(x.shape[-1], device=x.device, dtype=y.dtype).expand_as(y)
+        cans_addmm = torch.addmm if x.dim() == 2 else torch.baddbmm
+        for beta, alpha in _CANS_COEFFS:
+            p = z @ y
+            z = cans_addmm(z, p, z, beta=beta, alpha=alpha)
+            y = cans_addmm(y, y, p, beta=beta, alpha=alpha)
+
+        z = z.to(torch.float32)
+        z.mul_(torch.rsqrt(inf_norm))
+        return (z + z.mT) / 2.0
```
skyw: Nit: it is better to write *0.5. all compiler should be smart enough to do it. But doesn't hurt to do it explicitly.

tests/test_matrix_root_inverse_utils.py
```diff
@@ -0,0 +1,77 @@
...
+            device=FLAGS.device,
+        )
+
+        inverse_root = scaled_cans_coupled_ns(matrix)
+        identity = torch.eye(matrix.shape[-1], device=FLAGS.device)
+        whitened_matrix = inverse_root @ matrix @ inverse_root
+        matrix_root = torch.linalg.inv(inverse_root)
+        reconstructed_matrix = matrix_root @ matrix_root
+
+        torch.testing.assert_close(whitened_matrix, identity, atol=2e-4, rtol=2e-4)
```
skyw: Use `assert_close_to_identity` we added earlier.

tests/test_matrix_root_inverse_utils.py
```diff
@@ -0,0 +1,77 @@
...
+        previous_precision = torch.get_float32_matmul_precision()
+
+        inverse_root = scaled_cans_coupled_ns(matrix, fp32_matmul_prec=fp32_matmul_prec)
+
+        self.assertEqual(inverse_root.shape, matrix.shape)
+        self.assertEqual(inverse_root.dtype, torch.float32)
+        self.assertEqual(torch.get_float32_matmul_precision(), previous_precision)
+
+    def test_scaled_cans_inverse_root_accuracy(self) -> None:
+        matrix = torch.tensor(
```
skyw: parameterize couple of shapes, including 3d input.

emerging_optimizers/soap/matrix_root_inverse_utils.py
```diff
@@ -0,0 +1,101 @@
...
+    (2.410759275435182, -0.6186815444268036),
+    (2.1883348130094173, -0.5893091162177136),
+    (1.8595760874873613, -0.5449991062102938),
+    (1.589020160467417, -0.5075811685214573),
+    (1.5051653981684994, -0.4957799077972079),
+    (1.4925557853149838, -0.49259266842078675),
+)
+
+
+def scaled_cans_coupled_ns(
```
skyw: name doesn't seem quite reflect what it does.
How about `mat_root_inv_via_scaled_cans`? or similar

skyw:
@greptileai eps will never be negative. Python users are assumed consent adult.

skyw (dismissed):

skyw:
> Fair enough — if `eps` is always a small positive float by contract (e.g. `1e-6`), defensive validation is noise. The "consenting adults" principle is a reasonable Python convention.
> 
> Shall I remember this as a rule: _"Do not suggest adding defensive validation for internal numeric parameters (e.g., `eps`) where the caller contract already implies valid values"_?

@greptileai, yes, add that rule. Also update the confidence score of this PR with this rule in place.

skyw (approved):

PR #265, opened 2026-07-28
OKLS optimizer

mkhona-nvidia:
https://blog.tilderesearch.com/blog/online-kl-shampoo

### DIFF
--- a/docs/apidocs/soap.md
+++ b/docs/apidocs/soap.md
@@ -25,6 +25,17 @@ emerging_optimizers.soap
 .. autofunction:: update_eigenbasis_and_exp_avgs
 
 
+:hidden:`OKLS`
+~~~~~~~~~~~~~~~
+
+.. currentmodule:: emerging_optimizers.soap.okls
+
+.. autoclass:: OKLS
+    :members:
+
+.. autofunction:: update_kronecker_factors_okls
+
+
 :hidden:`REKLS`
 ~~~~~~~~~~~~~~~
 
@@ -38,5 +49,12 @@ emerging_optimizers.soap.soap_utils
 =====================================
 
 .. automodule:: emerging_optimizers.soap.soap_utils
+    :members:
+
+
+emerging_optimizers.soap.matrix_root_inverse_utils
+=====================================
+
+.. automodule:: emerging_optimizers.soap.matrix_root_inverse_utils
     :members:
 ```
--- a/emerging_optimizers/soap/__init__.py
+++ b/emerging_optimizers/soap/__init__.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from emerging_optimizers.soap.moso import MOSO
+from emerging_optimizers.soap.okls import OKLS
 from emerging_optimizers.soap.rekls import REKLS
 from emerging_optimizers.soap.soap import SOAP
 from emerging_optimizers.soap.stacked_soap import StackedSoap
@@ -20,6 +21,7 @@ from emerging_optimizers.soap.stacked_soap import StackedSoap
 
 __all__ = [
     "MOSO",
+    "OKLS",
     "REKLS",
     "SOAP",
     "StackedSoap",
--- a/emerging_optimizers/soap/matrix_root_inverse_utils.py
+++ b/emerging_optimizers/soap/matrix_root_inverse_utils.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/emerging_optimizers/soap/matrix_root_inverse_utils.py
+++ b/emerging_optimizers/soap/matrix_root_inverse_utils.py
@@ -19,7 +19,7 @@ from emerging_optimizers import utils
 from emerging_optimizers.utils import FP32MatmulPrecT
 
 
-__all__ = ["scaled_cans_coupled_ns"]
+__all__ = ["mat_root_inv_via_scaled_cans"]
 
 # All constant arithmetic, including the 1.01 safety factor, is folded into these coefficients.
 _CANS_COEFFS = (
@@ -36,7 +36,7 @@ _CANS_COEFFS = (
 )
 
 
-def scaled_cans_coupled_ns(
+def mat_root_inv_via_scaled_cans(
     x: Tensor,
     eps: float = 1e-12,
     fp32_matmul_prec: FP32MatmulPrecT = "highest",
--- a/emerging_optimizers/soap/matrix_root_inverse_utils.py
+++ b/emerging_optimizers/soap/matrix_root_inverse_utils.py
@@ -89,12 +89,14 @@ def scaled_cans_coupled_ns(
         if torch.get_float32_matmul_precision() == "medium":
             y = y.to(torch.bfloat16)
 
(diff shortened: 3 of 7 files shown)


xsgxlz:
Thanks for upstreaming OKLS! I opened a follow-up PR against your branch: [mkhona-nvidia/Emerging-Optimizers#2](https://github.com/mkhona-nvidia/Emerging-Optimizers/pull/2).

It adds:

- `mixed`, `high`, and `highest` scaled-CANS precision modes; `mixed` uses FP16 GEMMs with FP32 accumulation to avoid the NaNs observed with BF16 CANS, aligning with the Scaled CANS NS we proposed in the [blog](https://blog.tilderesearch.com/blog/online-kl-shampoo)
- Packed upper-triangular storage for symmetric Kronecker factors and inverse roots, reducing persistent symmetric-state memory by 50%

The changes passed 24 CUDA tests, Ruff checks, and multi-seed MLP trajectory comparisons against both the public release and internal implementation.

skyw:
> Thanks for upstreaming OKLS! I opened a follow-up PR against your branch: [mkhona-nvidia/Emerging-Optimizers#2](https://github.com/mkhona-nvidia/Emerging-Optimizers/pull/2).
> 
> It adds:
> 
> * `mixed`, `high`, and `highest` scaled-CANS precision modes; `mixed` uses FP16 GEMMs with FP32 accumulation to avoid the NaNs observed with BF16 CANS, aligning with the Scaled CANS NS we proposed in the [blog](https://blog.tilderesearch.com/blog/online-kl-shampoo)
> * Packed upper-triangular storage for symmetric Kronecker factors and inverse roots, reducing persistent symmetric-state memory by 50%
> 
> The changes passed 24 CUDA tests, Ruff checks, and multi-seed MLP trajectory comparisons against both the public release and internal implementation.

Thanks for the contribution. Our precision supports was mostly following what Pytorch would do for fp32 input. The bf16 cast was mainly because of pytorch lack of "medium" kernels. It would be a very significant testing cost if we expand beyond that and don't have plan to do so.

Packed storage is a great idea. Implementation will need more consideration as we'll need to incorporate rest of the code base as well as other NVIDIA SW stack's (megatron for example) requirement, which is out of this PR's scope.

skyw:
Also this PR was for demonstration and test.
We are changing abstraction of our shampoo/soap implementation (e.g. WIP https://github.com/NVIDIA-NeMo/Emerging-Optimizers/blob/skyw/okls_exp/emerging_optimizers/experimental/okls.py), so specific combination can be more contained and hopefully we have more room to accept things like fp16 support for just one version of OKLS.

xsgxlz:
> Also this PR was for demonstration and test. We are changing abstraction of our shampoo/soap implementation (e.g. WIP https://github.com/NVIDIA-NeMo/Emerging-Optimizers/blob/skyw/okls_exp/emerging_optimizers/experimental/okls.py), so specific combination can be more contained and hopefully we have more room to accept things like fp16 support for just one version of OKLS.

Thanks for the clarification. Because we found that, unlike Muon, bf16 almost certainly leads to a NaN for OKLS training, do you think we should remove the "medium" option? Or raise a warning for it?

xsgxlz:
Besides, we found that a tight estimate for the largest eigenvalue is helpful, especially when the model is large. Please consider this upper bound in [our repo](https://github.com/tilde-research/online-kl-shampoo-release/blob/main/okls/scaled_cans_coupled_ns.py) if it aligns with your requirements

```python
def _estimate_max_eigenvalue(A: torch.Tensor) -> torch.Tensor:
    """Strict upper bound via min(Wolkowicz-Styan, Minc-Sainte-Marie). Cost: O(n²)."""
    n = A.size(-1)
    diag = A.diagonal(dim1=-2, dim2=-1)

    # ── Optimal scaling: max(|A/c|) = √FP32_MAX / n ──
    # Both bounds involve O(n²·max²) intermediates.  Scaling by
    # c = max(|diag|)·n/√FP32_MAX keeps every intermediate in FP32 range
    # while maximising dynamic-range usage (fewest small elements lost).
    # For SPD matrices max(|A_ij|) ≤ max(diag_i), so this is a safe bound.
    _FROB_SCALE = n / math.sqrt(torch.finfo(torch.float32).max)  # n / 1.844e19
    c = (diag.abs().max(dim=-1).values * _FROB_SCALE).clamp(min=1e-30)
    c_inv = 1.0 / c

    # Single n² materialisation: |A/c|  (||x||_F = |||x|||_F, so both bounds
    # can work from the elementwise-abs form without loss of information.)
    abs_A_s = (A * c_inv[..., None, None]).abs()

    # ── Wolkowicz-Styan bound on A/c, then unscale ──
    m_s = diag.sum(dim=-1) * (c_inv / n)
    f_s_n = torch.linalg.matrix_norm(abs_A_s) * (1.0 / math.sqrt(n))
    s_s = torch.sqrt(torch.clamp((f_s_n + m_s) * (f_s_n - m_s), min=0.0))
    ws_bound = c * (m_s + s_s * math.sqrt(n - 1))

    # ── Minc-Sainte-Marie bound on A/c, then unscale ──
    d_s = torch.sum(abs_A_s, dim=-1)
    d_s_clamped = torch.clamp(d_s, min=1e-12)
    y_s = torch.einsum("...ij,...j->...i", abs_A_s, d_s_clamped)
    minc_bound = c * torch.max(y_s / d_s_clamped, dim=-1).values

    return torch.minimum(ws_bound, minc_bound)
```

skyw:
> > Also this PR was for demonstration and test. We are changing abstraction of our shampoo/soap implementation (e.g. WIP https://github.com/NVIDIA-NeMo/Emerging-Optimizers/blob/skyw/okls_exp/emerging_optimizers/experimental/okls.py), so specific combination can be more contained and hopefully we have more room to accept things like fp16 support for just one version of OKLS.
> 
> Thanks for the clarification. Because we found that, unlike Muon, bf16 almost certainly leads to a NaN for OKLS training, do you think we should remove the "medium" option? Or raise a warning for it?

Do you have a test case that CANS straight blow up (inf or nan) with bf16? That would be a sufficient reason to disallow bf16.
If it indirectly gets NaN deep in a training pipeline, probably default to TF32 and pop a warning for lower.

xsgxlz:
> > > Also this PR was for demonstration and test. We are changing abstraction of our shampoo/soap implementation (e.g. WIP https://github.com/NVIDIA-NeMo/Emerging-Optimizers/blob/skyw/okls_exp/emerging_optimizers/experimental/okls.py), so specific combination can be more contained and hopefully we have more room to accept things like fp16 support for just one version of OKLS.
> > 
> > 
> > Thanks for the clarification. Because we found that, unlike Muon, bf16 almost certainly leads to a NaN for OKLS training, do you think we should remove the "medium" option? Or raise a warning for it?
> 
> Do you have a test case that CANS straight blow up (inf or nan) with bf16? That would be a sufficient reason to disallow bf16. If it indirectly gets NaN deep in a training pipeline, probably default to TF32 and pop a warning for lower.

Yes - with BF16, the NS will almost surely explode for moderately large or ill-conditioned matrices. In our real training, we found BF16 survives at most 3 steps, so it might be better to disallow it.

```python
import torch

from emerging_optimizers.soap.matrix_root_inverse_utils import (
    mat_root_inv_via_scaled_cans,
)

device = "cuda"
size = 1024

generator = torch.Generator(device=device).manual_seed(0)
x = torch.randn(size, size, generator=generator, device=device)

matrix = (x @ x.T) / size
matrix.diagonal().add_(1e-6)

with torch.no_grad():
    for precision in ("medium", "high"):
        inverse_root = mat_root_inv_via_scaled_cans(
            matrix,
            fp32_matmul_prec=precision,
        )

        print(f"{precision}:")
        print("  finite:", torch.isfinite(inverse_root).sum().item())
        print("  NaN:", torch.isnan(inverse_root).sum().item())
        print("  Inf:", torch.isinf(inverse_root).sum().item())
```

```
medium:
  finite: 0
  NaN: 413232
  Inf: 635344
high:
  finite: 1048576
  NaN: 0
  Inf: 0
```

PR #269, opened 2026-07-29
ISO optimizer

mkhona-nvidia:
Implementation of ISO from https://arxiv.org/abs/2607.19331


emerging_optimizers/riemannian_optimizers/isospectral.py
```diff
@@ -0,0 +1,216 @@
...
+                self._init_state(param)
+
+                u = state["u"]
+                sigma = state["sigma"]
+                v = state["v"]
+                momentum_u = state["momentum_u"]
+                momentum_v = state["momentum_v"]
+
+                grad = param.grad
+                self._apply_weight_decay_inplace(param, grad, lr, weight_decay)
```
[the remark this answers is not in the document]
mkhona-nvidia: addressed by applying on sigma for all except L2

emerging_optimizers/rl_optimizers/isospectral.py
```diff
@@ -0,0 +1,245 @@
...
+        step_size: Optimizer step size.
+        retraction: Retraction method applied to both factors.
+
+    Returns:
+        The updated and retracted ``(u, v)`` factors.
+
+    Raises:
+        ValueError: If the retraction method is unsupported.
+    """
+    if retraction == "cayley":
```
skyw: Q: why not make it the same interface and combine with other retraction below?

emerging_optimizers/rl_optimizers/isospectral.py
```diff
@@ -0,0 +1,245 @@
...
+        if len(state) != 0:
+            return
+        if param.ndim != 2:
+            raise ValueError("ISO only supports 2D parameters")
+
+        # RL optimizer offload can place low-precision parameters on CPU, where
+        # linalg decompositions and retractions do not support float16 or bfloat16.
+        # Keep the factor and momentum state in FP32 while preserving parameter storage dtype.
+        factor_param = (
+            param.float() if param.dtype in (torch.float16, torch.bfloat16) else param
```
skyw: a bind param.float() should do, provide no one really use double.

emerging_optimizers/rl_optimizers/isospectral.py
```diff
@@ -0,0 +1,245 @@
...
+
+    @torch.no_grad()  # type: ignore[misc]
+    def _init_state(self, param: torch.Tensor) -> None:
+        state = self.state[param]
+        if len(state) != 0:
+            return
+        if param.ndim != 2:
+            raise ValueError("ISO only supports 2D parameters")
+
+        # RL optimizer offload can place low-precision parameters on CPU, where
```
skyw: Unnecessary comment, there may be other reasons it is in different dtype. Can comment why it must be float.

emerging_optimizers/rl_optimizers/isospectral.py
```diff
@@ -0,0 +1,245 @@
...
+                self._init_state(param)
+
+                u = state["u"]
+                sigma = state["sigma"]
+                v = state["v"]
+                momentum_u = state["momentum_u"]
+                momentum_v = state["momentum_v"]
+
+                # Match the gradient to the factor-state dtype used for low-precision CPU offload.
+                grad = param.grad.to(dtype=u.dtype)
```
skyw: Q: isn't u in "factor_param" dtype? always float?

emerging_optimizers/rl_optimizers/isospectral.py
```diff
@@ -0,0 +1,245 @@
...
+                state = self.state[param]
+                self._init_state(param)
+
+                u = state["u"]
+                sigma = state["sigma"]
+                v = state["v"]
+                momentum_u = state["momentum_u"]
+                momentum_v = state["momentum_v"]
+
+                # Match the gradient to the factor-state dtype used for low-precision CPU offload.
```
skyw: Removing mentioning of offloading, the code doesn't need to know and doesn't make assumption based on that. Besides, CPU offloading is not a standardized procedure.

emerging_optimizers/riemannian_optimizers/isospectral.py
```diff
@@ -0,0 +1,245 @@
...
+
+                u = state["u"]
+                sigma = state["sigma"]
+                v = state["v"]
+                momentum_u = state["momentum_u"]
+                momentum_v = state["momentum_v"]
+
+                # Match the gradient to the factor-state dtype used for low-precision CPU offload.
+                grad = param.grad.to(dtype=u.dtype)
+                if self.weight_decay_method == "l2":
```
skyw: This silently skip weight decay if method is not "l2", is that the intentional?

emerging_optimizers/rl_optimizers/isospectral.py
```diff
@@ -0,0 +1,245 @@
...
+                    if self.weight_decay_method != "l2":
+                        # Direct decay on param would be overwritten by the fixed-factor
+                        # reconstruction. Apply it to Sigma so the decayed spectrum persists.
+                        self._apply_weight_decay_inplace(sigma, sigma, lr, weight_decay)
+                    updated_param = (u * sigma.unsqueeze(0)) @ v.mT
+
+                state["u"] = u
+                state["v"] = v
+                state["step"] += 1
+                param.copy_(updated_param)
```
skyw: In general, avoid copy. can use addmm and set param to be output.

tests/test_isospectral.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def test_factor_state_is_orthonormal(self) -> None:
+        param = torch.nn.Parameter(torch.randn((7, 4), device=FLAGS.device))
+        optimizer = ISO([param], lr=1e-2)
+        param.grad = torch.randn_like(param)
+        optimizer.step()
+
+        state = optimizer.state[param]
+        identity = torch.eye(4, device=FLAGS.device)
+        torch.testing.assert_close(state["u"].mT @ state["u"], identity, atol=1e-5, rtol=1e-5)
```
skyw: use the identity check utility function.

tests/test_isospectral.py
```diff
@@ -0,0 +1,149 @@
...
+        torch.testing.assert_close(state["v"].mT @ state["v"], identity, atol=1e-5, rtol=1e-5)
+
+    @parameterized.named_parameters(
+        ("float16", torch.float16),
+        ("bfloat16", torch.bfloat16),
+    )
+    def test_cpu_low_precision_uses_fp32_factor_state(self, dtype: torch.dtype) -> None:
```
skyw: what's the reason to only test on CPU?

skyw (changes requested):

docs/apidocs/rl-optimizers.md
```diff
@@ -0,0 +1,10 @@
+```{eval-rst}
```
skyw: This feels weird, nothing in this optimizer is specific for RL. And RL doesn't have special requirment on how optimizer should be implemented.

Can mention in docstring that it is commonly used in what type of RL.

emerging_optimizers/riemannian_optimizers/isospectral.py
```diff
@@ -0,0 +1,245 @@
...
+        raise ValueError(f"Invalid retraction: {retraction}")
+
+    return (
+        retract(u - step_size * momentum_u),
+        retract(v - step_size * momentum_v),
+    )
+
+
+@registry.register_optimizer("iso")
+class Iso(opt_mixin.WeightDecayMixin, Optimizer):
```
skyw: Thanks. this looks much less itchy.

skyw (changes requested):

emerging_optimizers/riemannian_optimizers/isospectral.py
mkhona-nvidia: It doesn't skip weight decay, it applies the other methods directly on the singular values, since we reconstruct the param from it's singular value decomposition after every step (and hence applying on param would be a no-op, since it is reconstructed)

tests/test_isospectral.py
mkhona-nvidia: Replaced manual identity comparisons with `assert_close_to_orthogonal`.

tests/test_isospectral.py
mkhona-nvidia: Low-precision tests now use FLAGS.device, covering CPU and CUDA

docs/apidocs/rl-optimizers.md
mkhona-nvidia: It was designed specifically for RL, not to be used for pretraining

emerging_optimizers/rl_optimizers/isospectral.py
mkhona-nvidia: addmm cannot write to the param when there is a mismatch of precisions

docs/apidocs/rl-optimizers.md
skyw: Not enough to have a rl-optimizer catetory.

emerging_optimizers/riemannian_optimizers/isospectral.py
skyw: Where does it apply decay? if the if evaluate false, it skip `_apply_weight_decay_inplace`

emerging_optimizers/riemannian_optimizers/isospectral.py
```diff
@@ -0,0 +1,255 @@
...
+                    )
+                    if self.weight_decay_method != "l2":
+                        # Direct decay on param would be overwritten by the fixed-factor
+                        # reconstruction. Apply it to Sigma so the decayed spectrum persists.
+                        self._apply_weight_decay_inplace(sigma, sigma, lr, weight_decay)
+                    scaled_u = u * sigma.unsqueeze(0)
+                    if param.dtype == torch.float32:
+                        torch.addmm(param, scaled_u, v.mT, beta=0.0, out=param)
+                    else:
+                        # Mixed-dtype addmm cannot write an FP32 result directly into param.
```
skyw, line 230: That's a good concise comment.

docs/apidocs/rl-optimizers.md
skyw: Also the directory layout now I see it. Our directory are all optimizer algorithm oriented, not target application oriented.

emerging_optimizers/riemannian_optimizers/isospectral.py
```diff
@@ -0,0 +1,254 @@
...
+                        v,
+                        momentum_u,
+                        momentum_v,
+                        lr,
+                        retraction,
+                    )
+                    if self.weight_decay_method != "l2":
+                        # Direct decay on param would be overwritten by the fixed-factor
+                        # reconstruction. Apply it to Sigma so the decayed spectrum persists.
+                        self._apply_weight_decay_inplace(sigma, sigma, lr, weight_decay)
```
mkhona-nvidia: @skyw  here is where weight decay is applied

skyw: Ok, I see it now. I misread that both were == l2.

Logic is correct, but it is not the mixin was designed for exactly and very hard to read. Do you swith l2 and decoupled decay? or just use one?

docs/apidocs/index.md
```diff
@@ -13,4 +13,4 @@ psgd.md
 scalar-optimizers.md
 mixin.md
 utils.md
-```
\ No newline at end of file
+```
```
skyw: what was the change here? maybe an accidental end of line? looks harmless, but roll it back if you can.

emerging_optimizers/riemannian_optimizers/isospectral.py
mkhona-nvidia: in practice we only use decoupled. Added L2 for completeness

skyw (approved):

PR #283, opened 2026-08-10
Expose `Muon` parameters

janEbert:
It can be useful to obtain the `num_ns_steps` in certain cases, e.g., to balance compute around existing instances' `num_ns_steps`.

Could also refactor this to allow people to set `num_ns_steps` post-init by either (a) creating a new `scaled_orthogonalize_fn` when setting `num_ns_steps` or (b) simply passing `self` into the
`scaled_orthogonalize_fn` closure, so that the instance's current `num_ns_steps` are always accessed.

### DIFF
--- a/emerging_optimizers/orthogonalized_optimizers/muon.py
+++ b/emerging_optimizers/orthogonalized_optimizers/muon.py
@@ -91,6 +91,9 @@ class Muon(OrthogonalizedOptimizer):
             raise ValueError(f"num_ns_steps must be at least 1, got {num_ns_steps}")
 
         self._num_ns_steps = num_ns_steps
+        self._coefficient_type = coefficient_type
+        self._scale_mode = scale_mode
+        self._extra_scale_factor = extra_scale_factor
 
         if use_syrk:
             if torch.cuda.is_available():
@@ -106,6 +109,7 @@ class Muon(OrthogonalizedOptimizer):
                     sm_version,
                 )
                 use_syrk = False
+        self._use_syrk = use_syrk
 
         def scaled_orthogonalize_fn(grad: torch.Tensor) -> torch.Tensor:
             logging.debug(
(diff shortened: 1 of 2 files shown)


skyw:
> It can be useful to obtain the `num_ns_steps` in certain cases, e.g., to balance compute around existing instances' `num_ns_steps`.
> 
> Could also refactor this to allow people to set `num_ns_steps` post-init by either (a) creating a new `scaled_orthogonalize_fn` when setting `num_ns_steps` or (b) simply passing `self` into the `scaled_orthogonalize_fn` closure, so that the instance's current `num_ns_steps` are always accessed.



> It can be useful to obtain the `num_ns_steps` in certain cases, e.g., to balance compute around existing instances' `num_ns_steps`.
> 
> Could also refactor this to allow people to set `num_ns_steps` post-init by either (a) creating a new `scaled_orthogonalize_fn` when setting `num_ns_steps` or (b) simply passing `self` into the `scaled_orthogonalize_fn` closure, so that the instance's current `num_ns_steps` are always accessed.

It is very dangerous to modify `scaled_orthogonalize_fn` post initialization, so none of its arguments is maintained. What use cases do you have in mind, Jan? For simply use, inspect signiture can provide that information.

Having said that, we designed for relatively simple customization, like QKV split. If more complex customization is needed, we should revisit the entire design of OrthogonalizedOptimizer.

janEbert:
> It is very dangerous to modify scaled_orthogonalize_fn post initialization, so none of its arguments is maintained. What use cases do you have in mind, Jan? For simply use, inspect signiture can provide that information.

Enabling also setting `num_ns_steps` post-init was more a suggestion for potential build on top of this. Whether it's useful is up to you, but it isn't really related to this PR. It was just an idea that I could follow this PR up with.

For getting `num_ns_steps` from the closure: yes, it's possible with `inspect.getclosurevars`, but it's **super** hacky and I'd prefer to have a "simple" public-facing API to get the value.

The value is necessary to balance parallel work around the compute cost of the optimizer. I'm currently using this formula to approximate NS orthogonalization work:
```
cost = numel * (short_dim * num_steps + 1)
```

skyw:
> > It is very dangerous to modify scaled_orthogonalize_fn post initialization, so none of its arguments is maintained. What use cases do you have in mind, Jan? For simply use, inspect signiture can provide that information.
> 
> Enabling also setting `num_ns_steps` post-init was more a suggestion for potential build on top of this. Whether it's useful is up to you, but it isn't really related to this PR. It was just an idea that I could follow this PR up with.
> 
> For getting `num_ns_steps` from the closure: yes, it's possible with `inspect.getclosurevars`, but it's **super** hacky and I'd prefer to have a "simple" public-facing API to get the value.
> 
> The value is necessary to balance parallel work around the compute cost of the optimizer. I'm currently using this formula to approximate NS orthogonalization work:
> 
> ```
> cost = numel * (short_dim * num_steps + 1)
> ```

The design was against post initialization change of orthogonalization method, which would open too many doors and needs to be carefully guarded and tested. I don't have any case that we need to change that inflight yet.

Caller has all the information and cam maintain them if necessary. We can revisit when we have a case that Muon optimizer itself needs it. 

Alternative is storing and exposing all arguments to orthogonalization, which can be justified.

janEbert:
It's fair to not allow post-initialization changes.

As for the `num_ns_steps` information being available to callers: not necessarily. If I get a `Muon` optimizer instance passed from elsewhere and need to know its `num_ns_steps` as an outside observer, I currently cannot get this information.

What's the issue with adding the read-only attribute? Can I address it somehow?

skyw:
> It's fair to not allow post-initialization changes.
> 
> As for the `num_ns_steps` information being available to callers: not necessarily. If I get a `Muon` optimizer instance passed from elsewhere and need to know its `num_ns_steps` as an outside observer, I currently cannot get this information.
> 
> What's the issue with adding the read-only attribute? Can I address it somehow?

It's weird to have only `num_ns_steps` expose but not others, for example the coefficient type is highly coupled with number of steps. Understand calculating only steps matters for calculating cost.

I'd say do one of the following for short term:
* single private attribute `_num_ns_steps`, basically remove the property and setter guard. implies for internal (and adult) use only.
* make num_ns_steps, coefficient_type, scale_mode, extra_scale_factor property. basically saving everything need for orthogonalization.


Long term, there is a pending item on my end to make orthogonalizer class that should store those attribute. We started with very simple function but the level of customization we later added has beyond originally designed for .

janEbert:
Updated as per discussion, for now just similarly exposing the other parameters as read-only.

skyw (approved):

PR #287, opened 2026-08-14
Add vanilla shampoo

skyw:


emerging_optimizers/shampoo/soap_v3.py
```diff
@@ -445,7 +445,7 @@ def step(self, closure: Callable[[], float] | None = None) -> float | None:
 class KlSoapV3(SoapBase):
```
mkhona-nvidia, line 445: Can we use this PR to remove the "V3" terminology? 

It's what was used internally, but externally, we should just call it KlSoap and always have our best version (upto the limits of our understanding). We can document the assumptions in the docstring?

skyw: will do it with the klshampoo PR. This one has run CI too many times, and CI is busy recently.

PR #298, opened 2026-09-10
Add Sinkhorn functions

mkhona-nvidia:


emerging_optimizers/embedding_optimizers/__init__.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def _sinkhorn_balance(
+    update: torch.Tensor,
+    *,
+    eps: float,
+    num_steps: int,
+    zero_row_threshold: float,
+) -> torch.Tensor:
+    work = update.to(torch.float32)
```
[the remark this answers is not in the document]
mkhona-nvidia: EO embedding optimizers are intended for BF16/FP16/FP32 training; FP64 optimizer updates are not a supported use case. The FP32 working dtype is intentional. We can make that contract explicit with a dtype check or documentation rather than adding an FP64 execution path.

emerging_optimizers/orthogonalized_optimizers/sinkhorn_utils.py
```diff
@@ -0,0 +1,93 @@
...
+    # DeepSeek Algorithm 1 measures each row norm before balancing and masks rows whose norm is at most
+    # zero_row_threshold times the mean row norm. This keeps inactive or near-zero token rows at zero
+    # instead of amplifying their numerical noise during normalization.
+    row_norms = torch.linalg.vector_norm(balanced_update, dim=1, keepdim=True)
+    balanced_update.masked_fill_(row_norms <= zero_row_threshold * row_norms.mean(), 0.0)
+
+    # Apply the first row step separately, then express each remaining iteration as a column/row pair.
+    balanced_update.div_(row_norms.add_(eps))
+    for _ in range(num_steps // 2):
+        # Normalize columns along the row/token dimension.
```
skyw, line 83: remove not very meaningful comment, and potentially wrong if used in say some image model. code is clean enough.

emerging_optimizers/orthogonalized_optimizers/sinkhorn_utils.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # DeepSeek Algorithm 1 measures each row norm before balancing and masks rows whose norm is at most
+    # zero_row_threshold times the mean row norm. This keeps inactive or near-zero token rows at zero
+    # instead of amplifying their numerical noise during normalization.
+    row_norms = torch.linalg.vector_norm(balanced_update, dim=1, keepdim=True)
+    balanced_update.masked_fill_(row_norms <= zero_row_threshold * row_norms.mean(), 0.0)
+
+    # Apply the first row step separately, then express each remaining iteration as a column/row pair.
+    balanced_update.div_(row_norms.add_(eps))
+    for _ in range(num_steps // 2):
```
skyw, line 82: Q: why num_steps is set up this way? positive integer // 2?

I can see each loop normalize twice, but positive integer // 2 has no value in the code. num_xxx_steps would be a better name if num_steps to loop count may cause confusion.

tests/test_sinkhorn.py
```diff
@@ -0,0 +1,253 @@
...
+
+def setUpModule() -> None:
+    if FLAGS.seed is not None:
+        logging.info("Setting random seed to %d", FLAGS.seed)
+        torch.manual_seed(FLAGS.seed)
+        if torch.cuda.is_available():
+            torch.cuda.manual_seed_all(FLAGS.seed)
+
+
+def _sinkhorn_balance_reference(
```
skyw, line 37: AI copying code as reference? that's a bad reference in case both are wrong.
Implement reference in a different way, could be a very naive serial CPU version.

tests/test_sinkhorn.py
```diff
@@ -0,0 +1,253 @@
...
+        else:
+            column_norms = torch.linalg.vector_norm(result, dim=0, keepdim=True)
+            result = result / (column_norms + eps)
+
+    result = result * math.sqrt(update.size(1))
+    return result.to(update.dtype)
+
+
+class SinkhornBalanceTest(parameterized.TestCase):
+    def test_defaults_match_deepseek_v41(self) -> None:
```
skyw, line 61: Use inspect to check default value would be a better implementation for the purpose of this test.

tests/test_sinkhorn.py
```diff
@@ -0,0 +1,253 @@
...
+
+        torch.testing.assert_close(actual, expected, atol=0.0, rtol=0.0)
+
+    @parameterized.parameters(
+        ((8, 4), torch.float32, 1, 0.0),
+        ((9, 3), torch.float16, 3, 0.5),
+        ((17, 5), torch.bfloat16, 11, 1.0),
+        ((6, 6), torch.float32, 7, 0.0),
+    )
+    def test_matches_step_by_step_reference(self, shape, dtype, num_steps, zero_row_threshold) -> None:
```
skyw, line 75: if not exact match, test name should say "_close_to_"

tests/test_sinkhorn.py
```diff
@@ -0,0 +1,253 @@
...
+            rtol=1e-5,
+        )
+        torch.testing.assert_close(
+            result.square().mean(dim=0),
+            torch.ones(num_columns, device=FLAGS.device),
+            atol=0.05,
+            rtol=0.05,
+        )
+
+    def test_single_step_converts_unit_row_l2_norm_to_unit_rms(self) -> None:
```
skyw, line 112: Too many tests don't really add more protections. Use your best judgement to remove and keep the essential ones.

tests/test_sinkhorn.py
```diff
@@ -0,0 +1,253 @@
...
+        torch.testing.assert_close(actual, expected, atol=5e-7, rtol=5e-7)
+
+    def test_does_not_participate_in_autograd(self) -> None:
+        update = torch.randn((8, 4), device=FLAGS.device, requires_grad=True)
+
+        result = sinkhorn_balance(update)
+
+        self.assertFalse(result.requires_grad)
+
+    @parameterized.parameters(torch.float64, torch.int64, torch.complex64)
```
skyw, line 199: Nit: Combine input validation to the same test, otherwise it gets too noisy.

skyw (changes requested):

emerging_optimizers/orthogonalized_optimizers/sinkhorn_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    ``U[:, j] <- U[:, j] / (||U[:, j]||_2 + eps)`` for even steps.
+
+    The returned update is ``Delta = sqrt(n) * U``, where ``n`` is the number of columns. When no rows
+    are masked, this approximately targets ``(1 / n) * sum_j Delta[i, j]^2 = 1`` for every row and
+    ``(1 / m) * sum_i Delta[i, j]^2 = 1`` for every column, where ``m`` is the number of rows.
+
+    Args:
+        update: Dense signed update with shape ``(num_rows, num_columns)``.
+        eps: Numerical stability term added to each row or column norm.
+        num_steps: Total positive odd number of individual axis-normalization steps.
```
skyw: Nit: if we keep this, would be better to explain how steps are counted and why positive odd number here.

skyw (dismissed):

skyw (approved):

